PR #7, opened 2021-08-05
Add utilities to create mailto links

AnshG714:
Enable adding mailto links, which are generated based on pre-specified templates. 

How to use the templates
--
Each `.template` file will have 2 parts:
- The subject, wrapped between <subject> </subject> tags
- The body, wrapped between <body></body> tags. 

In a body tag, you can have have an arbitrary number of `<data>` tags. This is where the data passed into the `getMailUrl` function in `create-email-link.ts` will be inserted. 

Usage of `getMailUrl`
--
This is your contact point to get the mailto links with the correct recipient/subject/body email. Obviously, enter the receivers list and the template for which you want to generate an email href for (`type ValidTemplates = 'matched' | 'no-match-1' | 'no-match-2'`). The data is a string array which will directly map to the `<data>` tags you pass in. Note that the *number of <Data> tags and the length of the data array you pass in needs to be exactly the same*.

Now, in the actual email button, all you need to do is add an anchor tag in the following format: `<a href={getMailUrl(...)}> This is an email link </a>`. When you click the button, you should have your default mail client open up the correct email!!!

### DIFF
--- a/frontend/src/utils/emails/create-email-link.ts
+++ b/frontend/src/utils/emails/create-email-link.ts
@@ -1,6 +1,10 @@
-import { assert } from 'console'
 import parse from './template-parser'
 
+function assertEqual(expect: any, expression: any) {
+  if (expect !== expression) {
+    throw new Error(`AssertionFailure: expected ${expect}, got ${expression}`)
+  }
+}
 function constructMailtoUrl(
   receivers: string[],
   subject: string,
@@ -14,22 +18,24 @@ function constructMailtoUrl(
 }
 
 type ValidTemplates = 'matched' | 'no-match-1' | 'no-match-2'
-export default function getMailUrl(
+export default async function getMailUrl(
   receivers: string[],
   template: ValidTemplates,
   data: string[]
 ) {
   if (template === 'matched') {
-    assert(data.length === 0)
-    let { subject, body } = parse('matching.template', data)
+    assertEqual(0, data.length)
+    let { subject, body } = await parse('matching.template', data)
     return constructMailtoUrl(receivers, subject, body)
   } else if (template === 'no-match-1') {
-    assert(data.length === 2) // [student name, class]
-    let { subject, body } = parse('no-match-1.template', data)
+    assertEqual(2, data.length) // [student name, class]
+    let { subject, body } = await parse('no-match-1.template', data)
     return constructMailtoUrl(receivers, subject, body)
   } else if (template === 'no-match-2') {
-    assert(data.length === 1) // [student name]
-    let { subject, body } = parse('no-match-2.template', data)
+    assertEqual(1, data.length) // [student name]
+    let { subject, body } = await parse('no-match-2.template', data)
     return constructMailtoUrl(receivers, subject, body)
   }
 }
+
+getMailUrl([], 'matched', []).then((s) => console.log(s))
--- a/frontend/src/utils/emails/create-email-link.ts
+++ b/frontend/src/utils/emails/create-email-link.ts
@@ -1,0 +1,35 @@
+import { assert } from 'console'
+import parse from './template-parser'
+
+function constructMailtoUrl(
+  receivers: string[],
+  subject: string,
+  body: string
+): string {
+  let url = 'mailto:'
+  url += receivers.join(',')
+  url += '?subject=' + subject
+  url += '&body=' + encodeURI(body)
+  return url
+}
+
+type ValidTemplates = 'matched' | 'no-match-1' | 'no-match-2'
+export default function getMailUrl(
+  receivers: string[],
+  template: ValidTemplates,
+  data: string[]
+) {
+  if (template === 'matched') {
+    assert(data.length === 0)
+    let { subject, body } = parse('matching.template', data)
+    return constructMailtoUrl(receivers, subject, body)
+  } else if (template === 'no-match-1') {
+    assert(data.length === 2) // [student name, class]
+    let { subject, body } = parse('no-match-1.template', data)
+    return constructMailtoUrl(receivers, subject, body)
+  } else if (template === 'no-match-2') {
+    assert(data.length === 1) // [student name]
+    let { subject, body } = parse('no-match-2.template', data)
+    return constructMailtoUrl(receivers, subject, body)
+  }
+}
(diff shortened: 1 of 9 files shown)


frontend/src/utils/emails/create-email-link.ts
```diff
@@ -0,0 +1,35 @@
+import { assert } from 'console'
```
Coreball: Can't resolve module 'console'. Might be a part of Node

frontend/src/utils/emails/template-parser.ts
```diff
@@ -0,0 +1,70 @@
+// A naive template parser for the .template email files
+import { readFileSync } from 'fs'
+
+type EmailTemplate = {
+  subject: string
+  body: string
+}
+
+function readFile(filename: string) {
+  return readFileSync(`./bodies/${filename}`, 'utf-8')
```
Coreball: I don't believe `fs` is available for use on the client side. Though the type definitions exist, when actually trying to run the app this is not a function.

frontend/tsconfig.json
```diff
@@ -1,29 +1,21 @@
...
+    "target": "esnext",
+    "lib": ["dom", "dom.iterable", "esnext"],
     "allowJs": true,
     "skipLibCheck": true,
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
     "strict": true,
     "forceConsistentCasingInFileNames": true,
-    "module": "esnext",
+    "module": "commonjs",
```
Coreball, line 12: What are the changes to tsconfig for

Coreball (commented):
Good work, but I can't get it to run on my machine. I suspect it may be because certain functions are Node-exclusive and not available when running through the browser. How were you able to test it though?

frontend/src/utils/emails/create-email-link.ts
AnshG714: @coreball I basically tested it by running the typescript file as a script. That's what the changes in the ts-config are for.

Coreball: Oh ya that's why, runtime environments are different

frontend/src/utils/emails/create-email-link.ts
```diff
@@ -0,0 +1,41 @@
...
+    return constructMailtoUrl(receivers, subject, body)
+  } else if (template === 'no-match-1') {
+    assertEqual(2, data.length) // [student name, class]
+    let { subject, body } = await parse('no-match-1.template', data)
+    return constructMailtoUrl(receivers, subject, body)
+  } else if (template === 'no-match-2') {
+    assertEqual(1, data.length) // [student name]
+    let { subject, body } = await parse('no-match-2.template', data)
+    return constructMailtoUrl(receivers, subject, body)
+  }
```
Coreball, line 38: Maybe a switch statement could work better here, since `type ValidTemplates = 'matched' | 'no-match-1' | 'no-match-2'`. Or at least changing the last `else if` to an `else` so this function returns `string` instead of `string | undefined`.

frontend/src/utils/emails/create-email-link.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+getMailUrl([], 'matched', []).then((s) => console.log(s))
```
Coreball, line 41: Remove this line once we know it works

frontend/src/utils/emails/template-parser.ts
```diff
@@ -0,0 +1,69 @@
+// A naive template parser for the .template email files
+
+type EmailTemplate = {
+  subject: string
+  body: string
+}
+
+async function readFile(filename: string): Promise<string> {
+  return fetch(filename).then((r) => r.text())
```
Coreball, line 9: Good that `fetch` does work here! But the template files must be under `/public/` (not `/src/utils/emails/bodies/`)to be served by whatever's hosting the website

frontend/src/utils/emails/template-parser.ts
```diff
@@ -0,0 +1,69 @@
...
+  const lastIndex = text.indexOf(closeTag)
+
+  if (firstIndex === -1 || lastIndex === -1) {
+    throw new Error(`Invalid text provided -- could not parse ${tagName}.`)
+  }
+
+  return text.substring(firstIndex + tagName.length + 2, lastIndex).trim()
+}
+
+export default async function parseTemplate(
```
Coreball, line 57: Needing to read files/make network requests to generate a URL string does lead to some strange situations during testing:
```
const [url, setUrl] = useState('')
useEffect(() => {
  getMailUrl(['cl859@cornell.edu', 'ag759@cornell.edu'], 'matched', []).then((url) => setUrl(url!))
}, [])
```
to get `<a href={url}>` to work, because href can't be given a Promise directly. But it's possible this is less of an issue if email links aren't present on the page initially

Coreball (commented):
I am very slow at getting around to testing this but migration to fetch does work, though with some weird things. Would need to move the templates into the public folder if this is what we're gonna do

frontend/src/utils/emails/create-email-link.ts
AnshG714: I didn't use an `else` specifically because I didn't want to expose incorrect or under-development mailto links for prod. Maybe an AssertionError would be better suited here so we keep a contract, and then probably show an error toast on the frontend.

frontend/src/utils/emails/template-parser.ts
AnshG714: I think if we need to use fetch to get the templates from the public folder, then we will absolutely need Promises. A solution could just be to still display the link without any href attached to it, and then use react state + useEffect to dynamically add the mailto URLs. If this happens when the website loads, then this wouldn't be an issue.

frontend/src/utils/emails/create-email-link.ts
Coreball: All use of this function would be on the frontend though, right? So I don't think there's really an issue with exposing mailto links because all inputs to this would be controlled directly as part of the application (not user input). I don't think showing an error toast would be useful because its only purpose would be to show there's an internal development error on the frontend. Aka just don't call the function for templates that are not finished yet

frontend/src/utils/emails/template-parser.ts
Coreball: Hmmm... this feels like a lot of overhead for something is ultimately a long string with some substitution on the client side. Plus, this would be a lot of requests with the current structure if there's a link prepared for every grouping on the edit page and it needs to fetch every time.

Coreball:
Closing because assigned to @notwz now and may do it differently, maybe

PR #8, opened 2021-08-11
Resolve course name spacing issue

AnshG714:
this PR handles the spacing issue in course catalog names, as discussed before. The solution is to simply modify the roster API call to strip the excess spaces. This means we can now simply store whatever the client gives us into Firestore. Of course, _this means the client should be careful when sending course information_. We primarily use courseIds on the backend, not the names themselves, so whatever the client sends will be displayed on the course cards. 

Testing: 
```
>> getCourseId("CS 2110")
358546
>> getCourseId("ENGRD      2110")
358546
```

### DIFF
--- a/backend/functions/course/get_course_id.js
+++ b/backend/functions/course/get_course_id.js
@@ -30,7 +30,8 @@ function constructUrl(courseCatalogName, roster = "FA21") {
   return url;
 }
 
-async function getCourseId(courseCatalogName, roster = "FA21") {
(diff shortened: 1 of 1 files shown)


Coreball (approved):
Looks good to me. For reference, the validation snippet on the frontend is here in [Survey.tsx](https://github.com/cornell-dti/zing-lsc/blob/47a96a9fc19e0c296583a503ac35cd256b626b2a/frontend/src/modules/Survey/Components/Survey.tsx#L48)

Coreball:
Pls merge

PR #10, opened 2021-10-07
added "Miranda Luo" to README.md

mluo24:


Coreball (approved):
Consider keeping it in alphabetical role-separated order like the previous seasons (though that was a bit arbitrary when I wrote it. But following convention is good) Aka I listed devs first in the other sections

PR #12, opened 2021-10-19
Match button

mluo24:
From Notion:

> On the Edit Page, there should be a button at the top right which runs the LSC matching algorithm to generate study groups. As the algorithm is running, a popup + spinner + confirmation should appear (see designs for specifics). The Edit Page should be populated with the results when complete. 

This branch adds a match button and loading visuals (which Changyuan added), which makes a call to the "/make" endpoint, and then edits the functions that are behind the /make endpoint directly so that the correct information is returned to repopulate the state. This correct information are objects with a whole breadth of student information of groups and unassigned students as opposed to simple lists of objects/strings.


backend/functions/matching/functions.js
```diff
@@ -17,7 +18,7 @@ async function assertIsExistingCourse(courseId) {
 
 async function updateStudentReferencesForGroup(group, courseId) {
   return group.members.map((member) =>
-    updateStudentGroup(member, courseId, group.groupNumber)
+    updateStudentGroup(member.email, courseId, group.groupNumber)
```
Coreball: I had to set this to `member` instead of `member.email` to get it working because on line 127 the array being mapped is already `groupsWithMembersEmail`

backend/functions/matching/functions.js
```diff
@@ -86,42 +87,56 @@ async function makeMatches(courseId, groupSize = 2) {
...
   const matchedStudentSet = new Set(studentDataSliced.map((s) => s.email));
   const unmatched = unmatchedStudents.filter((s) => !matchedStudentSet.has(s));
 
+  // map for specific firebase data (just emails, as preserved from the previous function)
+  let groupsWithMembersEmail = groups.map((group) => {
+    return {
+      groupNumber: group.groupNumber,
+      members: group.memberData.map((member) => member.email),
+    };
+  });
```
Coreball: Could be const instead of let. Usually I make everything const unless it has to be reassigned because it conveys that nooo this isn't supposed to be changed. Helps prevent bugs. Also I personally prefer using parentheses to directly map to the object instead of a code block with a return statement, but it's ok if you leave it this way too.

frontend/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "frontend",
   "version": "0.1.0",
   "private": true,
+  "proxy": "http://localhost:5001/zing-lsc-backend/us-central1/api",
```
Coreball, line 5: Thanks for fixing this

frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -153,8 +157,39 @@ export const EditZing = () => {
...
+        setIsCurrentlyGrouping(false)
+      })
+      .catch((err) => {
+        console.error(err)
+        setShowError(true)
+      })
+    // old code
+    // new Promise((resolve) => setTimeout(resolve, 5000)).then(() => {
+    //   setIsCurrentlyGrouping(false)
+    // })
```
Coreball: Can delete this now

Coreball (approved):
Great work! I especially liked how you modified `makeMatches` to remember and return all the student data that it was already using to make matches instead of making a second API call or adding a call to `getStudentsForCourse` at the end of the match function. When I was testing I ran into an issue where `/make` would cause the loading to hang forever because `res.data.data` was undefined. Traced that back to the backend erroring and returning nothing in the response, which was caused by an error on the backend where `member.email` was undefined when updating student groups. The fix is pretty simple, I noted it in the comments below. Did you encounter this when testing on the latest commit of this PR? Once that's resolved, feel free to merge.

frontend/src/modules/EditZing/Components/EditZing.tsx
jessica0801: this code looks great! I like how you stayed consistent with your constant naming and how you used the is_ & set_ naming convention. That made the code very easy to follow!

backend/functions/matching/functions.js
jessica0801: I think what the coreball said is a good suggestion about using const instead of let just to make things more clear about what the restraints of this field is.

jessica0801 (commented):
Overall, I think you did an amazing job! The code is clean, and the variables are all well named!

PR #13, opened 2021-11-08
Create empty groups dashboard 

jessica0801:
I added the dashboard folder in the zing-lsc project with the components, styles, and type folders. 
The components folder includes dashboard, groupcard, and group files. 
The styles include a style file for dashboard, group card, and groups as well. 
Additionally, I also added the types index.file for getting course info for the dashboard once it has classes. 
Additionally, I added the index.ts file within the dashboard folder to export the dashboard.
Additionally, I connected the dashboard file to the /dashboard path so that it shows up when it runs on localhost:3000/dashboard.
I also changed the way that we exported the SVG files so that it is compatible with both google chrome and safari. 
I added the SVG files for the empty classes and moved all the code into the groups file so that it checks whether there are no groups made yet before it loads up the no classes to show page. 
Screen Shot 2021-11-20 at 10 27 32 PM


frontend/config-overrides.js
```diff
@@ -6,11 +6,13 @@ module.exports = override(
   addWebpackAlias({
     ['@assets']: path.resolve(__dirname, './src/assets'),
     ['@core']: path.resolve(__dirname, './src/modules/Core'),
+    ['@redux']: path.resolve(__dirname, './src/redux'),
```
Coreball: Might want to remove this for now, since we're not adding Redux yet and it be confusing

frontend/src/modules/Core/Constants/Theme.ts
```diff
@@ -3,8 +3,11 @@ export const colors = {
   verylightviolet: '#FCFAFF',
   black: '#222222',
   white: '#FFFFFF',
+  lightpurple: '#FBF9FF',
   darkpurple: '#CE9EF2',
+  purple: '#CE9EF2',
```
Coreball: This seems to be the same color as `darkpurple`, #CE9EF2

frontend/src/App.tsx
```diff
@@ -28,11 +30,13 @@ const App = () => {
         <Route exact path={SIGNUP_PATH} component={Signup} />
         <Route exact path={SURVEY_PATH} component={Survey} />
         <Route exact path={CREATE_ZING_PATH} component={CreateZingForm} />
-        <Route
+        <Route exact path={EDIT_ZING_PATH} component={EditZing} />
+        <Route exact path={DASHBOARD_PATH} component={Dashboard} />
+        {/* <Route
           exact
           path={`${EDIT_ZING_PATH}/:courseId`}
           component={EditZing}
```
mluo24: Is this still needed?

frontend/src/modules/Core/Constants/Routes.ts
```diff
@@ -4,8 +4,10 @@ export const SIGNUP_PATH = '/signup'
 export const SURVEY_PATH = '/survey'
 export const CREATE_ZING_PATH = '/createzing'
 export const EDIT_ZING_PATH = '/edit'
+export const DASHBOARD_PATH = '/dashboard'
 
 export const API_ROOT = 'http://localhost:5001/zing-lsc-backend/us-central1/api'
+export const USER_API = '/instructor'
```
mluo24: Why is this called USER_API and with a path of '/instructor'?

frontend/src/modules/Dashboard/Components/Groups.tsx
```diff
@@ -0,0 +1,65 @@
...
+          <StyledClassesContainer>
+            <StyledNoClasses />
+            <StyledNoClasses />
+            <StyledNoClasses />
+            <StyledNoClasses />
+            <StyledNoClasses />
+            <StyledNoClasses />
+            <StyledNoClasses />
+            <StyledNoClasses />
+          </StyledClassesContainer>
```
mluo24: There's too much repetition here. I feel like you should just loop this 8 times to clean it up.

frontend/src/modules/Dashboard/Components/Groups.tsx
```diff
@@ -0,0 +1,65 @@
...
+      )}
+      <StyledGroupArea>
+        <StyledGroupCardArea>
+          {groups.map((g, i) => (
+            <GroupCard
+              key={i}
+              id={g.courseId}
+              name={g.name}
+              submitted={300}
+              total={300}
```
mluo24, line 43: is this supposed to be hardcoded?

frontend/src/modules/Dashboard/Components/Groups.tsx
```diff
@@ -0,0 +1,65 @@
...
+  StyledNoClasses,
+} from 'Dashboard/Styles/Groups.style'
+
+import { GroupCard } from 'Dashboard/Components/GroupCard'
+import { CourseInfo } from 'Dashboard/Types'
+
+export const Groups = ({ groups }: ModalProps & GroupsProps) => {
+  return (
+    <StyledContainer>
+      {groups.length === 0 && (
```
mluo24, line 19: Good use of the inline conditional operator!

frontend/src/modules/Dashboard/Components/GroupCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+            }}
+            labelStyle={{
+              color: colors.white,
+              textTransform: 'none',
+              fontWeight: 600,
+              fontSize: '1rem',
+            }}
+            onClick={() => {
+              history.push(`/editZing/?id=${'z98mmggpO05931CviaUy'}`) // TODO: @shichong replace with real docID
+            }}
```
mluo24: I'm assuming this is almost exactly copy and paste from the groupeng counterpart, but is this still relevant?

mluo24 (commented):
Good job with this task! I am a bit unfamiliar with the specific task at hand, but I left some comments on parts that I believe are somewhat confusing or need clarification. In addition, before you push or anything, it's a good idea to get rid of the commented out code.

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "moment": "^2.29.1"
+  }
+}
```
mluo24: get rid of this (in the top level folder)

frontend/src/App.tsx
Coreball: The commented out version is actually the better implementation (allows using the useParams hook later in the edit page instead of extracting query params), not sure why line 33 was added

frontend/src/modules/Core/Constants/Routes.ts
Coreball: I think this was pasted from the other repository, but I agree that USER_API isn't as clear as something like INSTRUCTOR_API

frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -0,0 +1,112 @@
...
+  }
+  const handleClose = () => {
+    setAnchorEl(null)
+  }
+
+  const userEmail = 'cornell@cornell.edu'
+  // useAppSelector((state) => state.auth.user?.email)
+  const [groups, setGroups] = useState<CourseInfo[]>([])
+
+  var showMenu = false
```
Coreball: Unused variable

frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    setAnchorEl(event.currentTarget)
+  }
+  const handleClose = () => {
+    setAnchorEl(null)
+  }
+
+  const userEmail = 'cornell@cornell.edu'
+  // useAppSelector((state) => state.auth.user?.email)
```
Coreball: Probably just remove all Redux-related code since we're not using it yet and are not sure when it'll need to be added

frontend/src/modules/Dashboard/Components/GroupCard.tsx
```diff
@@ -0,0 +1,119 @@
+import React from 'react'
+import { useHistory } from 'react-router-dom'
+// import moment from 'moment'
```
Coreball: Delete

frontend/src/modules/Dashboard/Components/Groups.tsx
mluo24: Also I think it's worth noting that this doesn't preserve how many there are in each row, and I saw on the designs there are 4 of them in each row. Iteration through some list would also solve this issue (to have a fixed number of elements on each row). I'm not sure if this is intended behavior though, so if it is, ignore this comment.

frontend/src/modules/Dashboard/Components/GroupCard.tsx
Coreball: No, and I'd recommend just deleting unused code when copying over from the GroupEng side if it's not immediately used for the LSC product. This way when people work on this file next, it's more of a clean slate that has the minimum code necessary to function. So they don't get confused what this code is (I think the products are different enough that this code would not be shared).

frontend/src/modules/Dashboard/Components/Groups.tsx
```diff
@@ -0,0 +1,65 @@
+// export { }
```
Coreball: Remember to remove commented out code, we don't want it to be part of the app

frontend/src/modules/Dashboard/Components/Groups.tsx
Coreball: Haha I don't even remember what's going on with the GroupEng side for this

frontend/src/modules/Dashboard/Components/Groups.tsx
Coreball: I think this is screen-size dependent. With the current implementation it uses CSS flex and wrapping to add the second row based on how many can fit on one line

package.json
Coreball: Remove the package-lock.json too

Coreball (changes requested):
Good work porting over the dashboard code from the other repository! I know it can be a lot of code to handle at once. This feature will be very important to us with the Zing LSC product because it will be the way that LSC staff access the edit pages for the different courses they need to form groups for. I left more specific feedback in the review comments, but here are a few key suggestions:

- Generally, each PR should be able to "stand alone" in the sense that other devs would be able to pick up where you left off. This means the repository is clean, like removing commented code and unused variables (which can be confusing).
- I'd recommend not including code in the PR that has been copy pasted over and isn't immediately relevant to the product. This makes the code more confusing for developers because it's not being used yet. If we need that code later, we can always source it again.
- ^ it also makes it more complicated to review because the diff is larger and only some parts are related to the goal of the PR. PRs should be kept narrower in scope to the minimum code necessary.

Again, great work!

frontend/src/modules/Dashboard/Types/index.ts
```diff
@@ -0,0 +1,6 @@
+export interface CourseInfo {
```
Coreball: I think in other components this is not in a file named `index.ts`, rather that's used to re-export the content of the files

package.json
```diff
@@ -0,0 +1 @@
+{}
```
Coreball: Suggest to delete this file completely for now, along with package-lock.json

frontend/src/modules/Dashboard/Components/GroupCard.tsx
Coreball: Think u forgot this and a few other places (I don't think moment will be used here in the LSC product, since there are no deadlines for forms)

Coreball (approved):
Looks good other than a few minor things in comments! Just fix those and feel free to merge

PR #15, opened 2021-12-15
MUI v5 migration (from v4)

mluo24:
### Summary 

This pull request is the first step towards implementing the component library, where a good idea would be to update from the previous v4 to the new MUI v5 so that the component library would be MUI v5 based, and this would save the hassle of upgrading much later. 

I detailed most of the changes in the Notion page found [here](https://www.notion.so/51eb5ee682a04bd9bc88c88eb15ccb13?v=891947e574694c5ab6bf7c3fff0553f6&p=21bc5cfdd4ed43f9b84b052558cf098e) with the [migration guide](https://mui.com/guides/migration-v4/) found here, but the most important takeaways are detailed below:

Relevant Steps (all link to the upgrade guide):

- [Update React & TypeScript](https://mui.com/guides/migration-v4/#update-react-amp-typescript-version)
- [ThemeProvider setup](https://mui.com/guides/migration-v4/#themeprovider-setup)
- [Update MUI](https://mui.com/guides/migration-v4/#update-material-ui-version)
- [Run codemods](https://mui.com/guides/migration-v4/#run-codemods)
    - ran codemod `npx @mui/codemod v5.0.0/preset-safe <path>`
- [Handling Breaking Changes](https://mui.com/guides/migration-v4/#handling-breaking-changes)
- [Migrate from JSS](https://mui.com/guides/migration-v4/#migrate-from-jss)
    - ran codemod `npx @mui/codemod v5.0.0/jss-to-styled <path>`
    - the above command did not really help because some of the jss syntax was quite different from expected, so the codemod didn't like that. Most of this was done manually.

Breaking changes that have been fixed:

- span element that wraps children has been removed from the buttons, label classkey is also removed, so this has to be fixed (looking in the Button core component that is derived from a MUI button)
    - for this, I changed label to text, not sure if this will break anything
- Slider onChange (I think also onChange in general?) is now just type Event, seems to be the same with onClose as well in the snackbar (just had to change the type to an option type?): [https://mui.com/guides/migration-v4/#slider](https://mui.com/guides/migration-v4/#slider)
- `getContentAnchorEl` does not seem to exist anymore anywhere, so I just commented it usage of it out

Files changed:
- [x]  Button.tsx (sx)
- [x]  Checkbox.tsx (withStyles, other stuff)
- [x]  RadioButton.tsx (withStyles)
- [x]  InputField.tsx (theme, changed a lot of little things with previous JSS styling
- [x]  Slider.tsx (sx)
- [x]  Dashboard.tsx (adapt v4 Theme)
- [x]  MatchLoading.tsx (styled components)
- [x]  StudentGrid.tsx (styled components)

### Test Plan 

It's kind of hard to test this, just because the scope is big and that the only immediate and easy feedback is if there are any compilation errors. So, I have provided some screenshots below to see if they compare to the pre-migration stuff.

Screen Shot 2021-12-15 at 1 34 59 AM
Screen Shot 2021-12-14 at 4 50 12 PM
Screen Shot 2021-12-14 at 11 53 35 PM
Screen Shot 2021-12-15 at 1 35 19 AM

### Notes 

- The point of this PR is to immediately upgrade everything in the code base first to MUI v5, so any further implementations should be MUI v5 based. Therefore, this should be pushed first before the new component library updates that I will make in the future (effectively the end result would be the same but it would make a new PR much easier to handle).
- I really only changed the bare minimum so that the compiler won't be angry. Most of the changes in here does not necessarily mean they were good design changes, as I kept them as faithful as possible to the original implementation, which was not necessarily the best implementation choices either. This will likely be taken care of in future updates.
- The styled API here supports use of styled components, but I have not really modified anything so that that API would be taken advantage of with the various scattered styled components exported from some files.

### Breaking Changes 

- This is now using MUI v5. All instances of MUI v4 have been removed (or at least they should've been). After merging this in, everything should be based off of MUI v5 stuff.


frontend/src/modules/EditZing/Components/StudentGrid.tsx
```diff
@@ -1,37 +1,44 @@
...
-    },
-  })
+const PREFIX = 'StudentGrid'
+
+const classes = {
+  paper1: `${PREFIX}-paper1`,
+  paper2: `${PREFIX}-paper2`,
+}
+
+// should we make it so that when it overflows, then the user can scroll?
```
Coreball, line 19: Scrolling in a little box like this might look weird, not sure what a better solution is though

Coreball (approved):
Great work! Thanks so much for performing the migration from Material-UI v4 to MUI v5, it will let us take advantage of the improvements early on as we're migrating to the new design component system. I appreciate this being broken up into its own PR too, to make future changes easier to read and understand.

frontend/src/modules/EditZing/Components/StudentGrid.tsx
mluo24: Hmm, how should we fix this then? Also I'm not sure how it looks in the non upgraded version, because it seems like there is even less space after upgrading causing overflow. It also might be due to the fact that I'm on a Macbook Air currently so everything looks different.

Coreball: We should prob talk with design about it at some point, might involve using different breakpoints for the unmatched/group grids or using a min-width for the student card. Or maybe hiding overflow and showing all the details in a tooltip

mluo24: Yeah makes sense especially with the breakpoints, which hopefully should also be more fluid with the new component library, so should I just push what we got here for now?

Coreball: Yes

jessica0801 (commented):
Great job on the pull request, Miranda!! Thank you so much for initializing the change towards a more consistent styling! All the code seems very clean and consistent. This will really help us create more consistent styling and take advantage of all the pre-made components! Amazing job!!

PR #19, opened 2022-02-05
Standardized Component Library

mluo24:
## Summary 

This pull request aims to edit existing MUI components/create new components for the newly designed component library for Zing-LSC to be used in the application. The new/updated components can be viewed in http://localhost:3000/components (which should not be pushed to the main branch).

Notion page: https://www.notion.so/Standardize-Component-System-LSC-223c5a50014f4180889e2cf0e6acc18d

Figma: https://www.figma.com/file/5sae0s8rk6r9iVwpn74RY4/Zing-Components?node-id=12%3A0

Here is a list of the concepts/components from the Figma implemented (Typography is still in process), which were based on how reusable they are (i.e. not for a very specific application) and how much they resemble an original MUI component. Let me know if any of the unchecked should be implemented.

- [x] colors
- [ ] typography
- [ ] logo
- [x] all buttons
- [x] checkbox
- [x] radio buttons
- [x] dropdown
- [x] menus (I only did this for dropdown, should I be doing this for the whole Menu component?)
- [x] search bar
- [ ] filters
- [x] page numbers
- [ ] dividers
- [x] modal
- [x] progress bar
- [ ] "Class view"
- [ ] LSC-specific components

### How to use

- **Colors** are defined in the palette, and in certain situations (such as sx), 
- **Typography** is meant to be used with the Typography component, with the appropriate variants
- **Buttons, checkboxes, and radios** are all actual MUI components, and they should be used as such. However, for the secondary color, the "variant" prop must be set to "outlined" for it to work.
- **Dropdowns, search bar, page numbers, modal, and progress bar** are largely based off of (and are mostly directly heavily styled derivations of) similar MUI components, but they were edited heavily enough that they warranted separate components (is that a good idea?). Here is a small guideline for each of them:
    - The lists of props for each of them can be found in "frontend/src/modules/Core/Types/CustomComponent.ts". Almost all of them are actual MUI props that I believe would be relevant for all use cases.
    - **Dropdowns**: based off of the select component in MUI, which is based off of the native select tag in HTML/JSX. Props are available to control a bit of the sizing/style of the select, and it should be usable like the normal MUI select component, controlled or uncontrolled.
    - **Search Bar**: similar to dropdown
    - **Page Numbers**: this is based off of the pagination component, and the important props for the pagination component are available to use for this component. The most important change is `useRouter`, which is meant to change it so that it uses the react-router to navigate to a page, and it would append `?page={pageNumber}` to the end of the url.
    - **Modal**: this is based off of the dialog component. This is the most custom component, where it has the basic open, onClose, and children components, but it also has:
        - **title**: set the title of the modal (can leave blank)
        - **primaryButtonText**: the text for the primary (right) button in the modal, leave blank if you don't want the primary button
        - **secondaryButtonText**: same as above but for the left button
        - **primaryButtonProps**: MUI button props for primary button
        - **secondaryButtonProps**: same as above but for left button
    - **Progress Bar**: based off of linear progress. This is straight forward, where you can set a value explicitly from 0-100, or you set a fraction value explicitly (e.g. 1/8) and the value is calculated appropriately.

This PR does not implement these components in the app except for in Home, which is not actually part of the designs (I only changed it because the base button change messed up the custom component in the Core module), so I could revert that change if need be.

Let me know if I missed anything!

**###EDIT: 2/24**

sx support for custom components:
- DropdownSelect - sx is for the surrounding FormControl component, which to my knowledge can be used to edit the width, and inputSx is meant for the actual Select component
- PageNumbers - sx is for the full Pagination component, itemSx is for the PaginationItem component (the buttons that you can click on)
- ProgressBar - sx is for the LinearProgress (the same as the default MUI component)
- SearchBar - sx is for the TextField component, inputSx is meant for the Input component. The difference should be similar to the logic of the DropdownSelect
- ZingModal - sx is for the main Dialog component (same as MUI component). sx could also be applied to the buttons via the button props component. 

### Test Plan 

This is kind of hard to do for some of the non-MUI components. I have the components appearance-wise in Components.tsx available for viewing, but I did not test the full functionality of them/all potential use cases.

### Notes 

I am not sure if the way I did things is the best implementation, in terms of styling and file organization. I am also still not sure whether it is best practice to make an entirely new component customized or simply use style overrides, which I based it off of being how complicated editing the component would be/how specific the use case was. 

Some styles may be slightly off, and that was simple because they were not possible, or because it was not detailed in the Figma.

The responsiveness of the theme (mobile, tablet, etc.) is very questionable still. 

EDIT: 2/24
The support for sx on all of the custom components is now available. See above for how to use.

### Breaking Changes 

Not necessarily breaking changes, but I have modified some base components to the point that other components that are based off of them may be affected by those styles.


frontend/src/modules/Home/Components/Home.tsx
```diff
@@ -35,18 +36,17 @@ export const Home = () => {
...
-              label="Sign Up"
-              onClick={() => {
-                history.push('/signup')
-              }}
-            />
+            <Button
+              color="secondary"
+              variant="outlined"
+              component={Link}
+              to="/login"
```
Coreball, line 43: This is the better way of doing it, IMO

Coreball (approved):
This looks really good! Thanks for making these changes to implement the component system. I don't have many comments on implementation of specific elements, but in general I agree with your concerns about overriding styles vs new components. I think I slightly prefer overriding when possible, to keep the styling a thin layer over the original MUI component (and so we can reference the MUI docs more directly). But indeed some of these components are pretty complicated style-wise and that would be difficult. What you have is good, though I do wish the new components such as `ProgressBar` still had the `sx` prop available for specific customization. I don't think any of the unchecked items need to be implemented. Great work!

frontend/src/modules/Home/Components/Home.tsx
mluo24: Are you referring to how it was with "push history" before? lol

Coreball: New way is better

jessica0801 (commented):
Great job, Miranda! All the work and code looks really great, and the styling already looks so much cleaner. Everything seems to very consistent and well thought out. Thank you for taking the time to do all this!

Coreball (approved):
This is so incredibly cool. Being able to use sx props is nice. I think this will probably be the way we style things going forward.

Testing the new ProgressBar out briefly:

Old

https://user-images.githubusercontent.com/22627336/155654488-296c92cc-b0ea-4a9b-9548-92a780fab381.mov

New

https://user-images.githubusercontent.com/22627336/155654491-8ae42918-1d57-45a1-af40-6ccc31a26e62.mov

PR #21, opened 2022-02-13
Made student groups appendable

jessica0801:
### Bug Fix: Make the /students/{email} groups property appendable 

Previously, every time a student would fill out a new form, their groups field would be overwritten with the new class and a groupNumber  of -1. Now, the groups field is being appended to rather than overwritten so that it can store more than one class.

The current implementation of groups acts as a hashmap and appends a new courseId and groupNumber object when a student has not previously submitted a form for that class before. In other words, it will only add a class if the student does not already have that class in its groups field.

- now, when a student adds ["ENGRI 1270"] then ["ENGRI 1101"], their groups property will contain both 
- when a student has {courseId:"CS1110" groupNumber:2} in its groups field and submits another form for CS1110, it does not get rid of its current group.
- - when a student has {courseId:"CS1110" groupNumber:2} in its current groups field and submits the form for CS1112 and PHYS2213, it adds both these classes in with a default groupNumber of -1.
### Test Plan 
For my test plan, I was testing two different bugs. I was testing whether adding a new class would overwrite all previous information (so that the groups field always had length 1), and whether adding a course again would either overwrite or read-add course to the existing courses in groups.

- existing: {"CS 1110", -1} & add {"CS 1110", -1}, get only one in groups
- existing: {"CS 1110", 2} & add {"CS 1110", -1}, get only first one in groups
- existing: {"CS 1110", 2} & add {"PHYS 2213", -1}, get both in groups
- existing: {"CS 1110", 2} & add {"PHYS 2213", -1},{"CS 1110", -1} get {"CS 1110", 2} and {"PHYS 2213", -1} in groups.
### Breaking Changes 

- now,  /students/{email} groups property is appendable, and will be appended to no matter what class the student fill out the form with (even if the class is already in their groups array)


backend/functions/student/functions.js
```diff
@@ -36,26 +36,28 @@ const addStudentSurveyResponse = async (
...
     .set({
       name,
       college,
       year,
       preferredWorkingTime,
-      groups: crseIds.map((crseId) => ({
-        courseId: crseId,
+      groups:admin.firestore.FieldValue.arrayUnion( 
+      {
+        courseId: crseIds[0],
```
mluo24: Is there any particular reason you are only taking the first element of crseId?

mluo24 (approved):
Good job on this PR and taking on this bug fix! I think it's definitely important to have fixed it so students could add more classes. Your use of the firebase built-in function of arrayUnion was clever, so that you're just appending new (no duplicate) courses, but not overwriting previous ones.

You bring up not being able to request multiple courses at the same time? Is that why you are only taking the first courseId to merge into the array? It makes sense that it would be more complicated, but couldn't it also be a matter of combining the map from the previous version and then using the array union for each iteration?

Also, I'm not sure if your auto-formatting is working? It feels like it's deviating from what the formatting should look like with prettier.

You tested some good cases with your test plan too. Overall, great work!

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
Coreball: This and the other yarn file should not be in the pull request, you should probably remove them from your local repository too.

backend/functions/student/functions.js
Coreball: Only works for one element right now, the other crseIds are discarded

Coreball (commented):
Good job fixing this bug so far, it's something that's very easy to miss but will definitely be felt by users once the app is in production. I like how you consider the potential increased querying needed for fixing the bug. Your test plan is very useful as well!

The reference information for FieldValue is here: https://firebase.google.com/docs/reference/js/v8/firebase.firestore.FieldValue

Based on the function signature, I think it may accept multiple arguments, just as rest parameters instead of an array (because what if the user wants to union arrays of arrays). An array can be passed in using [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax).

For example:
```javascript
> console.log('hello')
hello
> console.log('hello', 'world') // taking two arguments
hello world
> console.log(['hello', 'world']) // logs an array
[ 'hello', 'world' ]
> console.log(...['hello', 'world']) // spread into separate arguments
hello world
```

@mluo24 I don't think auto-formatting is working because I think the precommit hook only applies to the frontend folder at the moment. But @jessica0801 still try to make it look like the rest of the code, e.g. a space after `groups:`, no trailing whitespace, `{ merge: true }`.

backend/functions/student/functions.js
```diff
@@ -19,7 +19,16 @@ async function removeStudentFromCourse(email, courseId, groupNumber) {
...
+//should_add compares inputted ids and existing ids to see which ones to add
+function should_add(ids,existing_set) {
+  const ids_to_add = new Array(); 
+  for (let i = 0; i < ids.length; i ++ ) {
+    if (!existing_set.has(ids[i])) {
+        ids_to_add.push(ids[i]);
+    }
+  }
+  return ids_to_add;
+}
```
Coreball: Sticking with existing naming convention in this file, we should use camelCase

backend/functions/student/functions.js
```diff
@@ -36,18 +45,26 @@ const addStudentSurveyResponse = async (
   const crseIds = await Promise.all(
     courseCatalogNames.map((name) => mapCatalogNameToCrseId(name))
   );
+  const ref = studentRef
+  .doc(email)
+  const existing_crses = (await ref.get()).data().groups;
```
Coreball: `.data()` is undefined when the document doesn't exist (i.e. this email has never submitted a survey), so getting groups from it throws an error

Coreball (changes requested):
Good work progressing on this bug fix! However, I have a few suggestions.

When I was testing the changes, I was unable to submit a survey as a new student who has never filled out the survey before. This is because calling `.data()` on a nonexistent document returns undefined. We need to account for the case where a new student is filling out the survey and doesn't have a student document yet.

I also believe there is a cleaner way to make the new group membership objects, perhaps something like this (note I did not test this):
```js
crseIds
  .filter(id => !existingIds.includes(id)) // don't add group again if it already exists
  .map(id => ({ courseId: id, groupNumber: -1 })) // map to group membership object
```

Naming convention things too. Again, good work! I hope to figure this out by next work session.

backend/functions/student/functions.js
```diff
@@ -36,18 +45,36 @@ const addStudentSurveyResponse = async (
   const crseIds = await Promise.all(
     courseCatalogNames.map((name) => mapCatalogNameToCrseId(name))
   );
+  var studentCrses = new Array();
```
Coreball: `var` is not good practice, use `let` instead if the variable needs to be re-assignable. https://github.com/cornell-dti/zing/pull/47#discussion_r736169822

backend/functions/student/functions.js
```diff
@@ -36,18 +45,36 @@ const addStudentSurveyResponse = async (
...
+    }))
+  }
+  catch(err){ 
+    crseIds.map((crseId) => {
+    studentCrses.push({
+      courseId: crseId,
+      groupNumber: -1
+    })
+   })
+  }
```
Coreball: I feel like wrapping the whole thing in a try-catch block is not the best way to approach this. We know exactly when to expect the document's data to be undefined, and can plan for it. Every new student will enter this catch block because when they submit the survey for the first time, their document doesn't exist yet.

```js
const existingData = (await studentRef.doc(email).get()).data()
const existingCourses = existingData ? existingData.groups : []
```

In this example, `existingCourses` becomes `existingData.groups` if available, otherwise `[]`.

Coreball (commented):
Thanks for continuing to work on this pull request. I did some testing on my own for the code I suggested in my previous review. Talk to me if you want to see what exactly I did, but it would eliminate the need for any for-loops or forEach calls, using map and filter only. I recommend taking that approach, as the code is much simpler, reducing the places where bugs can creep in. After the change I don't think `shouldAdd` would be needed anymore.

The general idea is pretty similar to what you have already:
1. Get existing data
2. Get existing courses, if available
3. Make new courses, which are the requested courses filtered to those that don't already exist
4. Combine existing courses and new courses

The **other important half of this bug** that I neglected to mention earlier, is with updating the courses. The code for that is just below the region you've been working in. Basically if a student is existing: {"CS 1110", 0} & add: {"CS 1110", -1} we get only the first one in the student's groups (which is what we want), but the **student email is added to the unmatched array in the course document even though the student is already in Group 0.** We can leverage the information we already have available to tell us which course documents we should update. If a student is already in a class, we don't need to add them to the course document again.

backend/functions/student/functions.js
```diff
@@ -56,8 +64,10 @@ const addStudentSurveyResponse = async (
       throw e;
     });
 
+  
   // Next, update each course record to add this student
-  const courseUpdates = courseCatalogNames.map((courseCatalogName, index) =>
+  const coursesToUpdate = courseCatalogNames.filter((crse) => crsesToAdd.includes(mapCatalogNameToCrseId(crse)))
```
Coreball: Try to avoid using `mapCatalogNameToCrseId` here, it's pretty expensive because it involves a network call for every catalog name

backend/functions/student/functions.js
```diff
@@ -37,17 +37,25 @@ const addStudentSurveyResponse = async (
...
+  //studentCourses becomes courseIds of existingData.groups if available, otherwise []
+  let studentCrses = existingData ? existingData.groups : []
+  const studentCrseIds = studentCrses.map((crse)=> crse.courseId)
+  const crsesToAdd = crseIds.filter((crse) => (!studentCrseIds.includes(crse)))
+
+  crsesToAdd.map((crse) => studentCrses.push({
+    courseId: crse,
+    groupNumber:-1
+  }))
+
```
Coreball, line 57: Personally I'd do something like the following to combine the existing courses and new courses, but what you have is acceptable too (but change it to `forEach` instead of `map`, since not using the return value)

```js
const studentCourses = [...existingCourses, ...newCourses]
```

Coreball (commented):
Yay nice progress! Just a few things now, also I didn't test it this time lol

Coreball (approved):
Nice job, I tested it and it appears to work as expected! Semicolon usage is a little inconsistent but we'll resolve that when we (hopefully) add a formatter to the backend.

backend/functions/student/functions.js
```diff
@@ -37,17 +37,32 @@ const addStudentSurveyResponse = async (
...
     .set({
       name,
       college,
       year,
       preferredWorkingTime,
-      groups: crseIds.map((crseId) => ({
-        courseId: crseId,
-        groupNumber: -1,
-      })),
+      groups: studentCrses
```
gracshan, line 65: This pull request looks great! I like how you condensed this block since you already know what each mapping will return ahead of time. The variable names are also well chosen to be self-explanatory, so great job on that!

PR #23, opened 2022-03-02
Continuing off from Ansh's Mailto [PR #7]

wizhaaa:
- change naming scheme
- change elifs to switch case statements
- move templates to public folder

### Summary 

See: https://github.com/cornell-dti/zing-lsc/pull/7 

Enable adding mailto links, which are generated based on pre-specified templates. 

How to use the templates
--
Each `.template` file will have 2 parts:
- The subject, wrapped between <subject> </subject> tags
- The body, wrapped between <body></body> tags. 

In a body tag, you can have have an arbitrary number of `<data>` tags. This is where the data passed into the `getMailUrl` function in `create-email-link.ts` will be inserted. 

Usage of `getMailUrl`
--
This is your contact point to get the mailto links with the correct recipient/subject/body email. Obviously, enter the receivers list and the template for which you want to generate an email href for (`type ValidTemplates = 'matched' | 'no-match-1' | 'no-match-2'`). The data is a string array which will directly map to the `<data>` tags you pass in. Note that the *number of <Data> tags and the length of the data array you pass in needs to be exactly the same*.

Now, in the actual email button, all you need to do is add an anchor tag in the following format: `<a href={getMailUrl(...)}> This is an email link </a>`. When you click the button, you should have your default mail client open up the correct email!!!

This pull request is the first step towards implementing feature Foo

- [x] implemented X
- [ ] fixed Y

### Breaking Changes 

### DIFF
--- a/frontend/src/utils/CreateEmailLink.ts
+++ b/frontend/src/utils/CreateEmailLink.ts
@@ -1,0 +1,48 @@
+import parse from './TemplateParser'
+export {}
+
+function assertEqual(expect: any, expression: any) {
+  if (expect !== expression) {
+    throw new Error(`AssertionFailure: expected ${expect}, got ${expression}`)
+  }
+}
+
+function constructMailtoUrl(
+  receivers: string[],
+  subject: string,
+  body: string
+): string {
+  let url = 'mailto:'
+  url += receivers.join(',')
+  url += '?subject=' + subject
+  url += '&body=' + encodeURI(body)
+  return url
+}
+
+type ValidTemplates = 'matched' | 'no-match-1' | 'no-match-2'
+
+export default async function getMailUrl(
+  receivers: string[],
+  template: ValidTemplates,
+  data: string[]
+) {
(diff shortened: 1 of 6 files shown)


frontend/src/utils/CreateEmailLink.ts
```diff
@@ -0,0 +1,48 @@
...
+
+function constructMailtoUrl(
+  receivers: string[],
+  subject: string,
+  body: string
+): string {
+  let url = 'mailto:'
+  url += receivers.join(',')
+  url += '?subject=' + subject
+  url += '&body=' + encodeURI(body)
```
b3njamint, line 18: I liked how you split the concatenations into several lines for readability. Maybe you don't need to do this 3 times though—you could combine it in one multi-line statement?

b3njamint (commented):
Good job so far! I looked at all of your changes and they looked good to me. I know you are still working on it so good luck and keep it up!

PR #24, opened 2022-03-02
Add timestamp for individual students and group matching

gracshan:
### Summary 
This pull request adds a timestamp to indicate when the student signed up on the website. This was done by adding a new field for a student on Firebase called "time", and utilizing the serverTimestamp function provided by Firestore to fetch the time when the request is sent. 

A timestamp was also added for group matching. Two new fields were introduced: one for when the group was created and another for when the group was last modified. 

An example of what this looks like in the database is below:
database

### Test Plan 
Testing was done using Insomnia. I used a POST request as I was trying to test whether the data was being sent to the server. An example of what the request looks like on Insomnia is provided below:
insomnia

For testing the group matching, I ran the Firebase Emulator and checked whether the time fields were showing up when a group was created or modified. 
Initially, there is no group field when none of the students have been matched yet:
Screen Shot 2022-03-18 at 9 54 07 PM
Screen Shot 2022-03-18 at 9 53 52 PM

When students are matched, the create and update time fields are set to the same timestamp:
Screen Shot 2022-03-18 at 9 54 17 PM
Screen Shot 2022-03-18 at 9 54 37 PM

When a group is modified (Student D is moved into the group), only the update time field is changed:
Screen Shot 2022-03-18 at 9 56 16 PM


mluo24 (approved):
Good work on adding a timestamp to the document in the Firestore! This feature applies important context to students filling out surveys. This is pretty much the canonical solution, and I don't think I would change it, except for maybe changing time to timestamp to really reflect what it means, thought not entirely necessary.

In your testing, I think it would also be a good idea to potentially make a get request (if there exists a route for it!) with the student's information so you could also see the received data from the request and check if the timestamp is there.

wizhaaa (commented):
Nice PR! I think this is about how I would also implement adding a time field to the firestore database. 

I agree with Miranda about testing a GET request to see if we can fetch the data as well. 

Also, depending on how we would want to use the time field, store it as either month and day or time or something if you need it. Or if there is an easy way to parse time in that format, you could leave it as is.

backend/functions/src/matching/functions.js
```diff
@@ -119,7 +121,12 @@ async function makeMatches(courseId, groupSize = 3) {
           .doc(courseId)
           .collection('groups')
           .doc(group.groupNumber.toString())
-          .set(group, { merge: true })
+          .set(
+            group,
+            { merge: true },
+            { createTime: admin.firestore.FieldValue.serverTimestamp() },
+            { updateTime: admin.firestore.FieldValue.serverTimestamp() }
```
Coreball: I don't think these two lines are necessary, I commented them out and it still works. Since this is JavaScript it probably ignores extra arguments passed to the `.set()` function instead of erroring

backend/functions/src/student/functions.js
```diff
@@ -65,6 +65,7 @@ const addStudentSurveyResponse = async (
       year,
       preferredWorkingTime,
       groups: studentCrses,
+      time: admin.firestore.FieldValue.serverTimestamp(),
```
Coreball: Agree with @mluo24 about the naming of this, now that other time-related properties are called `createTime` and `updateTime`. Maybe `timestamp` to make it more clear (since there's a `year` property for students too) or `surveyTime` or `submissionTime` to parallel how it's worded for groups.

Coreball (approved):
Good job Grace! Thanks for diving into the LSC backend to understand what each of the matching functions are used for and where we need to keep track of timestamps. In my testing it behaved exactly as expected, with students getting their survey submission time updated too when they submit the survey again. These changes will be very helpful for LSC in tracking when they make edits.

Approving this PR, but take a look at the comments I left. I think you can remove a few lines from a `.set()` and as the others suggested, make `time` more descriptive. After that, feel free to merge this PR! I am neutral on updating the JSON responses sent from the backend to include the timestamps. It will need to be done eventually, but it's probably OK to save them for a later PR since this one has been out for a while already and we can test by looking in the Firestore GUI.

@notwz I like this way of storing the timestamp, it's native to the database and includes more precision than month/day. In general I lean towards recording things in full precision because it's easier to extract month/day out of a full timestamp than the other way around.

PR #25, opened 2022-03-05
Add TypeScript and Prettier to backend

Coreball:
### Summary 

This pull request prepares the backend for TypeScript and installs Prettier, which is configured to run automatically.

- [x] added TypeScript support
- [x] added Prettier
- [x] configured Prettier to use same styles as frontend
- [x] configured Prettier to run on precommit hook using Husky

The backend directory has been reorganized to support TypeScript. Code present in backend/functions has been moved to backend/functions/src. When building the app using `npm run build` (runs tsc, the TypeScript compiler), emitted JavaScript appears under backend/functions/lib. The npm scripts have been updated to build before running the Firebase emulators. JavaScript source files are still allowed in the project to provide a path for migration.

Prettier has been added to provide a consistent code style with the frontend. The `.prettierrc.json` is the same in both locations. Prettier is installed in both the frontend and backend directories. It can be run with `npm run format`. It will also run automatically before each commit using Husky and git hooks. ESLint has been set up to handle linting only and will save the formatting for Prettier.

This pull request does not modify existing code, other than formatting it with Prettier. With the default ESLint configuration, the use of `require()` counts as an error. I downgraded it to a warning for now.

### Test Plan 

Cloned locally, ran `npm install`, made poor formatting decisions, was automatically fixed by Prettier when I committed. Changed some file extensions to `.ts` and had VSCode yell at me for things not having types.

### Notes 

Generally followed these resources:
- https://firebase.google.com/docs/functions/typescript
- https://prettier.io/docs/en/install.html
- https://typicode.github.io/husky/#/?id=manual

Ran into issues with Husky and git hooks. Husky operates significantly differently in v7 (latest) and v4 (used on frontend). It must run in the same location as the `.git` directory so `npm run prepare` goes back to the root and installs the hooks under backend/functions/.husky. In the actual hook we also need to `cd backend/functions` before running lint-staged.

The ESLint target is also es2019 because some newer features like `.flat()` on arrays are used in the code, which are not available in es2017, the default. However the Node engine set on this package is high enough that this shouldn't be an issue.

### Breaking Changes 

You will need to update the contents of your existing .env file to use `../` instead of `./` because `config.js` is one directory deeper.


mluo24:
When you say "You will need to update your existing .env file from ./ to ../ because config.js is one directory deeper." do you mean we need to move it to the same directory as config.js?

Coreball:
> When you say "You will need to update your existing .env file from ./ to ../ because config.js is one directory deeper." do you mean we need to move it to the same directory as config.js?

Update the text in the .env file to say ../something instead of ./something! Or you could move the .env file and the file it references to the same folder as config.js, that'd probably work too. I'll update the PR description to be more clear.

mluo24 (approved):
Nice job on putting Prettier and Typescript on the backend! Now the formatting can stay consistent and look as good as the frontend, and we don't have to deal with weird JavaScript quirks.

PR #31, opened 2022-03-19
Automated Emailing Feature Backend Implementation 

wizhaaa:
### Summary  

Using Nodemailer library to automate emails from the backend. 
All code change is in backend/functions. 

Run `npm i nodemaler` to install or `npm i` . 
Created a new folder `email` to store the emailing features where `routes` calls the `functions` which aligns with our current backend structure. 

Also used typescript and used new module import syntax in index.ts. Will work on updating the rest of index to updated modules syntax for next task. 

For the feature, there are multiple parameters when making a post request such as 

 { recipientsList, message, template, newStudent, existingStudents, className, customSubject, customBody  }  and may add or change when we implement the frontend portion of the code to see what exactly we need for it. 

This pull request is the first step towards implementing feature "Automated Emails" 

- [x] implemented backend function 
- [ ] need outlook app password 
- [ ] need to implement the front end 

### Test Plan 

Use insomonia to make post requests. 

Do `npm run emulate` in `backend/functions` 

POST at {{ _.baseUrl }}/email/send 
Where baseUrl = http://localhost:5001/zing-lsc-backend/us-central1/api 

image

Optional: Use Ethereal email testing suite to create a test emailing account as your user credientials. 
[]https://ethereal.email/

)### Notes

Need to add to the `.env` file the two variables: 

`EMAIL_PASS = "passwordhere"`
`EMAIL_EMAIL = "emailhere"` 

Can DM me for the actually credientials.


jewang25:
The code looks good so far. I like how the code for the emailing features stays consistent with the current backend structure. I also like how you included the test plan and used the Ethereal email testing suite to thoroughly test sending the emails.

backend/functions/src/email/functions.ts
```diff
@@ -0,0 +1,138 @@
...
+transporter.verify((err, succ) => {
+  if (err) {
+    console.log(' ❌ Transporter error: ❌ \n ', err)
+  } else {
+    console.log(
+      ' \n  ✅ Nodemailer transporter working ✅  \n Successful Message: ',
+      succ
+    )
+  }
+})
```
mluo24, line 27: I think you can get rid of these console.log statements, unless they serve a purpose I don't realize.

backend/functions/src/email/templates.ts
```diff
@@ -0,0 +1,257 @@
+const matched = () => {
+  return `Subject: Study Partners!
+  <div> 
+    Dear students,  
+  </div>
+    <br> 
+    <br>
```
mluo24, line 7: I'm speaking for this whole file, and I wanted to address this before, but is there a reason that everything is hardcoded? I feel like adopting a strategy with how the survey questions are made with JSON files as opposed to hardcoding (data-driven 😩) would be a way to go, especially with customization.

backend/functions/src/email/functions.ts
```diff
@@ -0,0 +1,138 @@
...
+    customSubject,
+    customBody,
+  } = req.body
+
+  const recipients = parseRecipients(recipientsList)
+  //  [body] can be of type email.
+  // {
+  //   matched, firstNoMatch, secondNoMatch, addStudent,
+  //   lateAddStudent, askJoinGroup, checkIn, custom
+  // }
```
mluo24, line 56: I would delete unused code comments

mluo24 (commented):
Good job on continuing what Ansh was doing before and getting it to work pretty well! Your testing plan seems pretty thorough especially with making requests and ethereal email. I left some comments specific to areas of your code. Also, you don't need to say to install specific packages in the PR, just `node install` is good enough since it's in the package.json (also you spelled "nodemailer" wrong lol)

backend/functions/src/email/functions.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+})
+
+const parseRecipients = (recipients: string[]) => {
+  let res = ''
+  recipients.forEach((e) => {
+    res += ', ' + e
+  })
+  return res
+}
```
Coreball, line 35: Would `recipients.join(', ')` work here?

backend/functions/src/email/functions.ts
```diff
@@ -0,0 +1,138 @@
...
+    // email,
+    recipientsList,
+    message,
+    template,
+    newStudent,
+    existingStudents,
+    className,
+    customSubject,
+    customBody,
+  } = req.body
```
Coreball, line 49: In the current backend we usually destructure the request body in routes.js and pass the values to the function in functions.js. I think that's preferable because it's more clear what the parameters for `sendEmail` are, as opposed to `req: any` and having to look at the implementation to see what `req.body` should contain.

Also why are name and email commented out but present in the testing screenshot?

backend/functions/src/email/functions.ts
```diff
@@ -0,0 +1,138 @@
...
+    to: string
+    cc: string
+    subject: string
+    html: string
+  }
+
+  const mail: mailType = {
+    from: ' "Cornell LSC Study Partners" <lscstudypartners@gmail.com>',
+    to: `jewell.schultz19@ethereal.email, wz282@cornell.edu`,
+    cc: `${recipients}`,
```
Coreball, line 118: During my testing the from: field doesn't actually use that's in the <>, it used the value of process.env.EMAIL_EMAIL which may not match with the from: field here

What are we going to use in the to: field later? Does LSC put all the recipients in the to: or cc:

backend/functions/src/email/functions.ts
```diff
@@ -0,0 +1,138 @@
...
+        msg: 'fail',
+      })
+    } else {
+      res.json({
+        msg: 'success',
+        email: message,
+      })
+      console.log('message sent with data being:: \n ', data)
+    }
+  })
```
Coreball, line 135: I think using async/await instead of passing `transporter.sendMail` a callback function is preferable. It would make it easier to handle req/res in routes.js like the other parts of the backend as well (this is related to an earlier comment about giving this function req vs all the parameters individually). Like return this promise and send a fail/success in routes.js. According to the nodemailer docs if you don't pass `sendMail` a callback it returns a promise object

backend/functions/src/email/templates.ts
Coreball: @mluo24 could do that, but then we have to find/replace magic values to substitute for variables instead of using string template literals. Good point tho

backend/functions/src/email/templates.ts
```diff
@@ -0,0 +1,257 @@
...
+    <br>
+
+    Happy studying!`
+}
+
+const firstNoMatch = (className: string) => {
+  return `< div > Dear Student, </>
+    < br >
+    <br>
```
Coreball, line 64: What is this HTML lol

backend/functions/src/email/templates.ts
```diff
@@ -0,0 +1,257 @@
...
+  const parseNames = () => {
+    let res = ''
+    studentsRaw.forEach((e: string) => {
+      res += ', ' + e
+    })
+
+    return res
+  }
+
+  const students = parseNames()
```
Coreball, line 141: Is this the same code as `parseRecipients`? See the comment there, also this results in `, a, b`

backend/functions/src/email/routes.ts
```diff
@@ -0,0 +1,13 @@
...
+import { config } from 'dotenv'
+import { sendEmail } from './functions'
+
+const router = Router()
+config()
+
+// Modulated sending email functions
+router.post('/send', (req, res, next) => {
+  sendEmail(req, res)
+})
```
Coreball, line 11: From what I see, the basic structure here in the other parts of the backend is something like
1. extract values from `req.body`
2. call function with the values
3. use returned promise to `res.status`

Coreball (commented):
Thanks for working on automated emailing! This is a critical step for the product before launch, and will make it much easier for LSC to integrate the product into their workflow. I have left several comments, but overall I applaud your ability and initiative to develop this new feature on your own. Good job!

PR #32, opened 2022-03-19
Google Sign-In Design Implementation

mluo24:
### Summary 

This PR is a followup to https://github.com/cornell-dti/zing-lsc/pull/30

This implements the design at https://www.figma.com/file/9zueLCvKT7BBMYi8Dt8Ukp/Zing-LSC?node-id=2010%3A168025 while also attempting to use less of excessive styled-components styling and more sx and native MUI styling.

### Test Plan 

I compared the design on various (computer) screen sizes to the Figma design.
image
image

**EDIT: MODIFIED**

After speaking with Becky and Alyssa, it was determined that the mobile design does not matter, so designing for mobile is negligible. I modified some of the background:

Fitting to the screen on the Figma 1440 x 900
image

Fitting to my screen (Macbook Pro 14 inch)
image

The background is set to 100% for both directions, so they stretch to fit the screen. It isn't an issue visually since they are all SVGs, but it does cause noticeable change in position when switching screen sizes. However, it does not seem to be distracting. I believe this is a good compromise for the lack of designs for different screen sizes.

### Notes 

To use the `inheritViewBox` property for the `SvgIcon` component, I upgraded MUI to version 5.5.1. So far, there were no breaking changes.

~~Also note that the PR contents of https://github.com/cornell-dti/zing-lsc/pull/30 are also present here.~~

Never mind, I did a bunch of rebasing and force pushing and now everything is gone.


b3njamint (commented):
I looked through all of your code and everything is written very well. Good job!! I like how you cleaned up the previous code and deleted the commented-out code to make everything more readable. I also like how you implemented the history stack too.

Looking at the screenshot of your screen, it seems to be a little different from the design. For example, you are missing the bottom right blob, and a lot of the blobs seem to be kind of high. I don't know how much difference is ok, but I just wanted to let you know what I noticed.

frontend/package.json
```diff
@@ -22,6 +22,7 @@
     "@types/react-dom": "^17.0.0",
     "axios": "^0.21.1",
     "date-fns": "2.11.1",
+    "firebase": "^9.6.8",
```
b3njamint: Nothing is broken by specifying the firebase version, right?

mluo24: I don't think so. `yarn install` for firebase automatically put this value in the package.json.

CollinWoo (commented):
Great work! The implementation looks really close to the original figma design and the code runs great on my computer. I also like how you removed some of the commented out code that wasn't being used; it makes the programs look a lot cleaner.

But I noticed that when resizing the screen, some of the text started to have some weird positioning and sizing and the sign in button started to drift out of its container:
Screen Shot 2022-03-23 at 5 54 34 PM
Screen Shot 2022-03-23 at 5 55 01 PM

I don't think this application is going to be used on especially small screens so this issue is very minor. Otherwise, everything looks great!

frontend/src/modules/Home/Components/Home.tsx
```diff
@@ -24,19 +28,29 @@ export const Home = () => {
     <StyledBackground>
       <StyledContainer>
         <StyledLeftPanel>
-          <StyledLogo />
           <StyledWhiteActionText>
-            Create groups, connect students
+            Connect students, create Zings
```
Coreball: Is this the text we want to have in the end? I thought we were deprecating calling things "Zings"

Coreball (approved):
Looks good! Pretty wild to resize the window and watch the blobs grow and shrink and move around.

frontend/src/modules/Home/Components/Home.tsx
mluo24: Not sure, that was just on the Figma. I can ask Becky about it.

PR #35, opened 2022-03-30
Added Authorization for Requests on Backend

mluo24:
### Summary 

This pull request is almost the last step in the authentication journey for Zing-LSC. This resolves the security problem of attempting to make requests to routes for the app without being authenticated. Middleware was added so authentication is required (having an authtoken in the header) for the route to be accessible. In the frontend, axios receptors were used so that the header when authenticated could be added automatically when a user makes a request. Then the app actually make these requests when the user is logged in.

The only backend route not protected is for submitting the survey (anyone can do that).

### Test Plan 

image
image
image

I tested on Postman with a protected route on the backend three different situations: no authtoken, an incorrect authtoken, and a correct one. All three cases produced the right results, with a 403 forbidden on the first one and a success on the last one. 

The only unprotected route (posting for the survey) works. 

Also, everything works when using the app through the frontend, which is the most important part (no unauthorized errors in the console when trying to get stuff with axios). 

### Notes 

When you want to make requests to the backend now, you now need to add a header variable called "authtoken" and set it to the `idToken` to be able to get through to the requests. I'm not sure how to do that other than `console.log` in App.tsx.

### Breaking Changes 

A bunch of rebases were made. Lowkey it broke everything (on GitHub), but I am not sure. Let me know if it does.

### DIFF
--- a/backend/functions/src/middleware/auth-middleware.js
+++ b/backend/functions/src/middleware/auth-middleware.js
@@ -3,15 +3,19 @@ import * as admin from 'firebase-admin'
 // function courtesy of https://itnext.io/how-to-use-firebase-auth-with-a-custom-node-backend-99a106376c8a
 // checks if valid based on the auth token in firebase admin
 export function checkAuth(req, res, next) {
-  if (req.headers.authtoken) {
+  // Bearer token
+  if (req.headers.authorization) {
     admin
       .auth()
-      .verifyIdToken(req.headers.authtoken)
+      .verifyIdToken(req.headers.authorization.split(' ')[1])
       .then(() => {
         next()
       })
-      .catch(() => {
-        res.status(401).send('Unauthorized: error occurred')
+      .catch((err) => {
+        console.log(err)
+        res
+          .status(401)
+          .send('Unauthorized: error occurred (credentials may be wrong)')
       })
   } else {
     res.status(401).send('Unauthorized: please sign in')
--- a/backend/functions/src/middleware/auth-middleware.js
+++ b/backend/functions/src/middleware/auth-middleware.js
@@ -1,9 +1,19 @@
-let authorized = false
+import * as admin from 'firebase-admin'
 
+// function courtesy of https://itnext.io/how-to-use-firebase-auth-with-a-custom-node-backend-99a106376c8a
+// checks if valid based on the auth token in firebase admin
 export function checkAuth(req, res, next) {
-  if (authorized) {
-    next()
+  if (req.headers.authtoken) {
+    admin
+      .auth()
+      .verifyIdToken(req.headers.authtoken)
+      .then(() => {
+        next()
+      })
+      .catch(() => {
+        res.status(403).send('Unauthorized: error occurred')
+      })
   } else {
-    res.status(403).send('Unauthorized')
+    res.status(403).send('Unauthorized: please sign in')
   }
 }
--- a/backend/functions/src/middleware/auth-middleware.js
+++ b/backend/functions/src/middleware/auth-middleware.js
@@ -11,9 +11,9 @@ export function checkAuth(req, res, next) {
         next()
       })
       .catch(() => {
-        res.status(403).send('Unauthorized: error occurred')
+        res.status(401).send('Unauthorized: error occurred')
       })
   } else {
-    res.status(403).send('Unauthorized: please sign in')
+    res.status(401).send('Unauthorized: please sign in')
   }
 }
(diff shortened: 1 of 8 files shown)


backend/functions/src/middleware/auth-middleware.js
```diff
@@ -0,0 +1,19 @@
+import * as admin from 'firebase-admin'
+
+// function courtesy of https://itnext.io/how-to-use-firebase-auth-with-a-custom-node-backend-99a106376c8a
```
Coreball: Thanks for citing, easy to see where you got the code from

backend/functions/src/middleware/auth-middleware.js
```diff
@@ -0,0 +1,19 @@
...
+      .auth()
+      .verifyIdToken(req.headers.authtoken)
+      .then(() => {
+        next()
+      })
+      .catch(() => {
+        res.status(403).send('Unauthorized: error occurred')
+      })
+  } else {
+    res.status(403).send('Unauthorized: please sign in')
```
Coreball: I think [401 Unauthorized](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) is the right status code compared to [403 Forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403)

frontend/src/App.tsx
```diff
@@ -19,17 +19,42 @@ import { useEffect, useState } from 'react'
...
+          })
+          .catch(() => {
+            setIsLoading(false)
+          })
+      else {
+        setIsLoading(false)
+      }
     })
-  }, [])
+  }, [currentUser, token])
```
Coreball: I thought we left it as `[]` because onAuthChanged binds a callback function that runs whenever auth state changes already?

frontend/src/App.tsx
```diff
@@ -19,17 +19,42 @@ import { useEffect, useState } from 'react'
...
+            axios.interceptors.request.use(
+              (request) => {
+                request.headers.authtoken = idToken
+                return request
+              },
+              (error) => {
+                return Promise.reject(error)
+              }
+            )
+            setToken(idToken) // might have to leave this in here
```
Coreball: Wait why do we need this, I tried removing it and the useState for token and it still appears to work fine for me?

backend/functions/src/student/routes.js
```diff
@@ -22,15 +23,15 @@ router.post('/survey', (req, res) => {
   )
     .then(() => res.status(200).json({ success: true }))
     .catch((err) => {
-      if (err.name == 'processing_err') {
-        return res.status(500).send({ sucess: false, message: err.message })
+      if (err.name === 'processing_err') {
+        return res.status(500).send({ success: false, message: err.message })
```
Coreball, line 27: Thanks for fixing spelling lol

Coreball (commented):
Great work diving into something less familiar and adding authorization for requests on the backend! Following your test instructions, I also got the expected results. I appreciate the attention to detail adding middleware to the DELETE student as well, even though we haven't used that route yet. This PR is a major step for making our app more secure.

As for the authorization header, I think the standard way of doing it in this case is with a Bearer token (see [Fireship example](https://fireship.io/snippets/express-middleware-auth-token-firebase/) but basically just with how the auth token is formatted as a header) (more details on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization)).

Changing it to a bearer token also integrates nicely with Postman, since setting the authorization automatically generates the header:
Screen Shot 2022-04-02 at 19 34 34
Screen Shot 2022-04-02 at 19 34 48
Similarly in Insomnia (using a variable)
Screen Shot 2022-04-02 at 19 35 14

Here's what it looks like in the web console after the change (note the little question mark like the other headers). The network tab of the web console is also a nice way to get the token for testing without needing to `console.log`!
Screen Shot 2022-04-02 at 19 39 22

Again, great work!

frontend/src/App.tsx
mluo24: Initially, for some reason, the first login would always not be authorized (return a 403), and I wasn't sure why. I tried a bunch of random combinations and this one was the only one that worked consistently, but that being said, it is one of said random combinations, so I guess I can start process of elimination to see which parts are actually needed.

frontend/src/App.tsx
mluo24: I guess this is one of the ones not needed lol

Coreball (approved):
TypeScript 👍

PR #37, opened 2022-03-31
(attempt 3) evolving to ts backend 

wizhaaa:
### Summary

<sub>hopefully this pr isn't branching off my old mailto pr for some reason</sub>

Refactoring the backend into Typescript with modern imports and exports. 

- [x] functions.js -> functions.ts
- [x] routes.js -> routes.ts
- [x] lots of types are just any 
- [ ] Should change the types into their actual type when you use the functions as I don't really know all of them explicitly and don't want to break the code right now. 

### Test Plan 

Ran some tests through the majority of all the routes in Insomnia. They be working so far. 

### Notes 

Should change the types into their actual type when you use the functions as I don't really know all of them explicitly and don't want to break the code right now.


backend/functions/src/config.ts
```diff
@@ -0,0 +1,13 @@
+import admin = require('firebase-admin')
```
Coreball: `import admin from 'firebase-admin'` worked for me

backend/functions/src/config.ts
```diff
@@ -0,0 +1,13 @@
...
+import { config } from 'dotenv'
+config()
+
+const serviceAccount: string = require('../service-account.json')
```
Coreball, line 5: I guess if we're hardcoding the SA path we don't need to config anymore

backend/functions/src/course/get_course_id.ts
```diff
@@ -30,7 +30,7 @@ function constructUrl(courseCatalogName, roster = 'FA21') {
   return url
 }
 
-async function getCourseId(courseCatalogNameRaw, roster = 'FA21') {
+async function getCourseId(courseCatalogNameRaw: string, roster: any) {
```
Coreball: Can probably say `roster: string` because with the default argument from before, we know it's a string like `'FA21'`

backend/functions/src/matching/functions.ts
```diff
@@ -188,15 +206,15 @@ async function addUnmatchedStudentToGroup(courseId, groupNumber, studentEmail) {
 
 // transfer student from group1 to group2
 async function transferStudentBetweenGroups(
-  courseId,
-  group1,
-  group2,
-  studentEmail
+  courseId: string,
+  group1: { toString: () => string },
+  group2: { toString: () => string },
```
Coreball: This looks very weird

Coreball (approved):
Thanks for these changes, it's an important step towards adopting TypeScript on the LSC backend. Understandably most types are still left as `any` which isn't particularly helpful at the moment, but it would be a lot of work for one PR to figure out the types for everything. Good job!

PR #38, opened 2022-04-04
Deploy Firebase Functions and Firebase Hosting

Coreball:
### Summary 

This pull request sets up deployment of the backend to Firebase Functions and the frontend to Firebase Hosting. This is done with the new Firebase project `zing-lsc`, which supersedes `zing-lsc-backend` as the Firebase project for development. The new project is because the old project was under the cornell.edu organization, rendering deployed functions unable to invoked without special authorization.

There is only one `firebase.json` located at the root, instead of one located at [the root](https://github.com/cornell-dti/zing/blob/360236715a21310b555581d4f3e81413180e92a2/firebase.json) and one located under [backend](https://github.com/cornell-dti/zing/blob/360236715a21310b555581d4f3e81413180e92a2/backend/firebase.json) like in Grouper. There should only be one because this is the same project, and it moves us closer to a good monorepo instead of two separate projects sharing the same repository.

GitHub integration has been set up to deploy Firebase Hosting preview channels on new pull request. It should also deploy to the live channel (main URL) when PRs are merged. Not able to test this yet, but it's set up on Grouper.

The frontend will use the local backend URL when developing and the deployed backend URL when deployed. Added the `cors` package to the backend to allow any origin. This also removes the need to proxy during frontend development, so it has been removed (thanks @mluo24 for testing)

- [x] updated to use new Firebase project
- [x] manual deploy backend to Firebase Functions
- [x] manual deploy frontend to Firebase Hosting
- [x] automatic deploy frontend to Firebase Hosting with [Firebase Hosting GitHub Action](https://github.com/marketplace/actions/deploy-to-firebase-hosting)

### Test Plan 

Ran `npm run deploy` in backend/functions and frontend to run my new scripts. For the frontend this will build and then deploy, for backend it will build as part of a predeploy hook.

Visited https://zing-lsc.web.app and confirmed that it sends requests to https://us-central1-zing-lsc.cloudfunctions.net/api (deployed URLs).

Visited http://localhost:3000 when running locally and confirmed that it sends requests to http://localhost:5001/zing-lsc/us-central1/api (development URLs).

Sent requests to https://us-central1-zing-lsc.cloudfunctions.net/api via Insomnia, worked as expected.

Cloned repository on `deploy-dev` branch and ran `npm install` for backend/functions, made a test commit to confirm that Husky is still working. Initially it wasn't, which prompted https://github.com/cornell-dti/zing-lsc/commit/f7affc7275227fe156f6c4c2faa0bb27bf16fe2f to fix it for both deployment and local development. Putting it in regular dependencies isn't optimal but can't figure out how to ignore the `prepare` script when deploying. It failed because Husky was a devDependency and isn't installed with `npm ci`.

~~Unable to test deployment via GitHub Pull Requests because this PR needs to be merged first.~~

This pull request was deployed with the GitHub Action successfully

### Notes 

Modified `firestore.rules` to block all reading and writing. This is the default for Firestore, and we don't do any frontend database access. All database access is currently done with the Admin SDK.

### Breaking Changes 

Need to use a new `service-account.json` to have the backend use the new Firebase project. The frontend changes to use the new Firebase project have been committed to version control.


hexbabe (approved):
Went over comments verbally during Team Time

- CORs concern (not an issue due to auth)
- Firebase jargon review

PR #39, opened 2022-04-16
Linked dashboard and edit page through view button

jewang25:
### Summary 

Used react useHistory to hook up dashboard and edit page through view button so that the edit page displays the correct course based on the course's id that is in Firebase. 

### Test Plan 

- Test with multiple classes
- Test with different users taking different classes

### Notes 

- Example: http://localhost:3000/edit/358546, where 358546 is the course's id in Firestore.


wizhaaa (commented):
Nice work! Pretty quick implementation and it works on my end as well.

gracshan:
Great job implementing this feature! I like how concise you kept it. It works for me!

CollinWoo:
Works perfectly for me! I like how your compact code makes the implementation very readable and how you deleted the commented out code. Great job!

frontend/src/modules/Dashboard/Components/GroupCard.tsx
```diff
@@ -22,11 +22,17 @@ function Alert(props: any) {
 
 export const GroupCard = ({
   key,
+  id,
   name,
   newStudents,
   groupsFormed,
 }: GroupCardProps) => {
-  // const history = useHistory()
+  let history = useHistory()
```
Coreball: Since `history` will not be reassigned later, it's safer to use `const` instead of `let` because accidentally reassigning the variable will show an error. I use `const` by default everywhere, and use `let` if I need the variable to be reassignable.

Coreball (approved):
Good work! This is pretty much the ideal way to do it, it's very clear and straightforward to understand. Linking the dashboard and edit page is critical to making our app actually usable.

PR #40, opened 2022-04-17
Add tooltips for groups and students

gracshan:
### Summary 

This pull request implements tooltips for students and groups. When the mouse is hovered over the group name, the time that it was created appears as text underneath. For students, the hover triggers a text that shows the time they requested a study partner.

### Test Plan 
I used Insomnia to populate the database with sample students and form a group. Then I started up the frontend to test whether the hover displays the correct contents.

For students, the submissionTime field was used.
Screen Shot 2022-04-17 at 11 31 02 AM
Screen Shot 2022-04-17 at 11 41 12 AM

For groups, the createdTime field was used.
Screen Shot 2022-04-17 at 11 41 36 AM
Screen Shot 2022-04-17 at 11 41 57 AM

### Notes 

- The date is formatted as MM/DD/YYYY
- The group time shown is the time that it was created (not modified)


jessica0801 (commented):
Great job on this PR! I think you did an amazing job employing features such as mapping to create clean and concise code. Also, great job using Tooltips to implement this feature, I am very excited about it

frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -47,7 +48,12 @@ export const EditZing = () => {
       .get(`${API_ROOT}${COURSE_API}/students/${courseId}`)
       .then((res: AxiosResponse<CourseStudentDataResponse>) => {
         setUnmatchedStudents(res.data.data.unmatched)
-        setStudentGroups(res.data.data.groups)
+        setStudentGroups(
+          res.data.data.groups.map((group) => ({
+            ...group,
+            createTime: new Date(group.createTime),
+          }))
+        )
```
mluo24: I like your use of the spread operator!

frontend/src/modules/EditZing/Components/StudentGrid.tsx
```diff
@@ -65,7 +69,17 @@ export const StudentGrid = ({
...
       <div ref={drag}>
         <Paper style={{ opacity: opacity }} className={classes.paper1}>
-          {student.name}
+          <Tooltip
+            disableFocusListener
+            disableTouchListener
+            title={
+              /* Could use getMonth and getDate to format mm/dd */
+              'Requested study partner ' +
+              new Date(student.submissionTime).toLocaleDateString()
```
mluo24: Is `student.submissionTime` not of type Date already?

frontend/src/modules/EditZing/Styles/StudentAndGroup.style.tsx
```diff
@@ -59,3 +59,9 @@ export const StyledStudentDetail = styled.p`
   margin: 0;
   padding: 0;
 `
+
+export const StyledStudentText = styled.p`
+  font-weight: 800;
+  margin: 0;
+  padding: 0;
+`
```
mluo24, line 61: I recommend you make use of MUI v5 styling as opposed to more styled-components bloat.

frontend/src/modules/EditZing/Components/GroupGrid.tsx
```diff
@@ -33,7 +35,16 @@ export const GroupGrid = ({
         style={{ opacity: isOver ? '0.6' : '1' }}
       >
         <StyledGroupTextWrapper>
-          <StyledGroupText>{'Group ' + String(groupNumber)}</StyledGroupText>
+          <Tooltip
+            disableFocusListener
+            disableTouchListener
+            title={
+              /* Could use getMonth and getDate to format mm/dd */
+              'Created on ' + createTime.toLocaleDateString()
```
mluo24: Do you know what exact format you need? I believe that if you want mm/dd exactly you should change it to what you have in your comment, but also I'm not sure if this edge case would occur (most of this would happen in the US I would assume).

mluo24 (commented):
Good work on this feature! The code is cleanly written and I could understand what you were trying to do with each line, from grabbing the data from the database to displaying it. I'm not approving this yet as when I tried to run it, it breaks with the current database data, as the current data does not have the new fields (getting undefined values when you query the data, I'm assuming from submissionTime), so I think you should handle this case or fix the database to include this.

jewang25 (commented):
The code looks good! I like how you made the code easy to understand like naming the variables nicely. The feature also looks really nice and useful.

frontend/src/modules/EditZing/Components/EditZing.tsx
Coreball: Though we aren't using the value yet it'd be good to convert `updateTime` to a Date here as well, so the object conforms to the correct TS type.

frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -21,6 +21,7 @@ import {
 import { API_ROOT, COURSE_API, MATCHING_API } from '@core/Constants'
 import { useParams } from 'react-router-dom'
 import { MatchLoading } from './MatchLoading'
+import { group } from 'node:console'
```
Coreball: Perhaps an accidentally added line

frontend/src/modules/EditZing/Components/StudentGrid.tsx
Coreball: For consistency, it would be better to convert `submissionTime` to a Date object when the data is retrieved instead of when it is displayed. This would involve mapping `res.data.data.unmatched` for unmatched students and each group's `memberData` for the grouped students.

Coreball (changes requested):
Great work on the frontend side of tooltips and using MUI! I am personally excited for this feature as I love keeping track of timestamps and when things happened. The screenshots you provided are very descriptive of what exactly is being implemented as well.

I ran into one major issue while testing. The match button works, but when the frontend receives the newly matched data back it crashes. I suspect this is because the dates in the groups and students aren't converted when [processing the new groups](https://github.com/cornell-dti/zing-lsc/blob/56b6f1a826c5f7ab5e60d044a12f2a6c7cb1cc0a/frontend/src/modules/EditZing/Components/EditZing.tsx#L172-L179), but also because the backend doesn't (a) convert student submission times to Dates before sending and (b) send back group creation/update time when running [makeMatches](https://github.com/cornell-dti/zing-lsc/blob/56b6f1a826c5f7ab5e60d044a12f2a6c7cb1cc0a/backend/functions/src/matching/functions.ts#L61). Everything works fine when refreshing the page though, because the groups are fetched the normal way. It breaks when pressing the match button as the response for that contains the new groups, which are not handled yet.

Data currently returned from `/matching/make` (when the green match button is pressed), note that submissionTime is still stringified database-native Timestamp instead of Date and there are no create time or update time fields for the group:

Screen Shot 2022-04-18 at 23 00 40

Let me know if I can help in any way, looking forward to this!

Coreball (approved):
Looks really good! The issue with the match button has been fixed and everything works as expected in my testing. Good job!

PR #42, opened 2022-04-17
Restrict Permissions

mluo24:
### Summary 

This pull request is (maybe) the final step for implementing the LSC authentication, where this part would be restricting page viewing permissions for the accounts who are not authorized to view the pages (everyone else who is not on Zing or in LSC). If an account who is not on the whitelist logs in, all they will see is a page indicating they are not authorized.

**EDIT (5/9/2022)**: This PR now has a decent amount of changes from the last few commits. I updated the boolean logic to use string literals as types, which simplifies the code a lot and got rid of a lot of pesky bugs (thanks @Enochen). I added a loading screen to the Dashboard here as part of the implementation for the unauthorized and authorized states (also fixing a lot of errors being thrown in the console), though now it's not really necessary and just a nice touch.

### Test Plan 

I just logged in with an approved and non approved account to make sure that everything could either be accessed or not accessed. 

If not approved:

image

### Notes 

~~This is an issue more of with the Google authentication via Firebase and the axios interceptors: when testing out switching accounts to see if the middleware works, a bug was encountered where the token was not refreshed unless the page was hard refreshed when switching accounts, so if the previous account could access the page, logged out, and a new account that can’t access it logs in, then that account could access until a page refresh. This was discussed and deemed not necessarily an issue, as if someone were to log in and log out and log into a different account, it would likely be a person who would've had access anyway.~~

~~In addition, the logout of the 403 error page seems to only work when clicked twice (probably having to do with the redirect stuff going on).~~

All should now be fixed (thanks @Coreball)!

Potentially sus way of handling the endpoint for checking if a user is authorized.

### Breaking Changes 

Added a new collection called "allowed_users" in Firestore.


backend/functions/src/middleware/auth-middleware.ts
```diff
@@ -23,3 +41,39 @@ export function checkAuth(req: Request, res: Response, next: NextFunction) {
...
+    admin
+      .auth()
+      .verifyIdToken(idToken)
+      .then((decodedToken) => {
+        const uid = decodedToken.uid
+        admin
+          .auth()
+          .getUser(uid)
+          .then((user) => {
+            if (user.email) {
```
b3njamint: could these nested if-else statements be simplified into just: 
```
if(user.email && allowedUsers.includes(user.email)) {
    next()
} else {
    res.status(403).send('Unauthorized: not correct permissions')
}
```

b3njamint (commented):
Good job Miranda on your long authentication journey—it was very inspirational seeing the constant iterations and direction changes that you went through. On this PR specifically, I really liked how you implemented the whitelist. I didn't notice any errors (as far as I can tell).

Coreball (approved):
Looks good! This addition is very important for keeping our data secure. I've mentioned this elsewhere already (saying it again here for completeness) but it would be very helpful to have the add-authorized-users script work with the local Firestore emulator or to have a pre-populated database (perhaps by [importing data on start](https://firebase.google.com/docs/emulator-suite/connect_firestore#import_and_export_data)) so people can still use the `npm run emulate` script easily. That is, having the emulated database start off with the authorized users in it so we don't have to add it manually every time when testing.

backend/functions/src/index.ts
```diff
@@ -18,14 +26,20 @@ const app = express()
...
+app.get('/getauth/:email', checkAuth, (req, res) => {
+  const allowedUsersRef = db.collection('allowed_users')
+  allowedUsersRef.get().then((r) => {
+    const email = req.params.email
+    const data = r.docs.map((doc) => doc.data().email)
+    res
+      .status(200)
+      .send({ success: true, data: { isAuthed: data.includes(email) } })
+  })
+})
```
Coreball: A way you could make this more secure would be to use the request token (which is already included by the time this endpoint is accessed, plus `checkAuth` is there already) to get the email like you do with `checkIsAuthorizedHelper`, because currently this allows any signed in user to query if someone else is an administrator or not, instead of limiting it to self-checks

frontend/src/auth/AuthContext.tsx
```diff
@@ -1,9 +1,21 @@
...
+ * 'loading' = state is currently loading (i.e. requests are currently being made, auth state currently unknown)
+ * 'unauthenticated' = you are not logged in
+ * 'unauthorized' = you are logged in but unauthorized/not in the allowed_users db
+ * 'authorized' = you are authorized/in the allowed_users db/have full access to the application
+ */
+export type AuthState =
+  | 'loading'
+  | 'unauthenticated'
+  | 'unauthorized'
+  | 'authorized'
```
Coreball, line 14: This is very nice, and I like how you added descriptions of what each state represents!

frontend/src/firebase/firebase.js
```diff
@@ -17,5 +17,5 @@ export async function signInWithGoogle() {
 }
 
 export async function logOut() {
-  signOut(auth)
+  signOut(auth).then(() => {})
```
Coreball: Is this `.then(() => {})` necessary?

frontend/src/modules/Dashboard/Components/GroupCard.tsx
```diff
@@ -20,7 +20,7 @@ function Alert(props: any) {
 }
 
 export const GroupCard = ({
-  key,
+  // key,
```
Coreball: Could just delete it

Coreball (approved):
Very good job Miranda! By restricting permissions on the backend and showing unauthorized messages on the frontend, this keeps the data secure while also providing a good user experience for non-authorized users. The Firestore emulator import is very helpful.

frontend/src/firebase/firebase.js
mluo24: No, but WebStorm got mad at me so I just put it in there to appease it. I'll take it out bc I don't like the way it looks

Coreball (approved):
Good

PR #43, opened 2022-04-21
Email sent property

mluo24:
### Summary 

This pull request implements part 1 of the frontend email task for "newly matched", which essentially adds a new field in the database for a timestamp of when the email for "newly matched" has been sent (either the time or a null field for it hasn't been sent). The frontend takes this value in a request and displays a purple dot accordingly for whether the email has been sent or not. If the purple dot is there, that means the email has been sent and the timestamp has a value, and hovering over it will reveal the type of email sent and the date it was sent out.

### Test Plan 

This was tested by adding data manually and testing out the two states: a null timestamp field or a filled in timestamp field (done manually). The match button was also pressed again and users were moved between groups to ensure that those functionalities would not break it.

A null timestamp:
image

A non-null timestamp with the dot being hovered over, showing tooltip: 
image

### Notes 

This PR does NOT add a feature to actually be able to change the field (i.e. this does not actually implement the email sending), so it must be changed manually in firebase. There will likely be merge conflicts with https://github.com/cornell-dti/zing-lsc/pull/40

### Breaking Changes 

- Database schema change: added a new field called shareMatchEmailTimestamp to the group schema.

### DIFF
--- a/frontend/src/modules/EditZing/Components/EditZing.tsx
+++ b/frontend/src/modules/EditZing/Components/EditZing.tsx
@@ -165,9 +165,13 @@ export const EditZing = () => {
       .post(`${API_ROOT}${MATCHING_API}/make`, { courseId: courseId })
       .then((response) => {
         let newGroups = studentGroups.concat(response.data.data.groups)
-        console.log(response.data.data.unmatched)
+        newGroups = newGroups.map((group: Group) => ({
+          ...group,
+          shareMatchEmailTimestamp: group.shareMatchEmailTimestamp
+            ? new Date(group.shareMatchEmailTimestamp)
+            : null,
+        }))
         setUnmatchedStudents(response.data.data.unmatched)
-        console.log(newGroups)
         setStudentGroups(newGroups)
         setIsCurrentlyGrouping(false)
       })
@@ -202,6 +206,7 @@ export const EditZing = () => {
               key={index}
               studentList={studentGroup.memberData}
               groupNumber={studentGroup.groupNumber}
+              shareMatchEmailTimestamp={studentGroup.shareMatchEmailTimestamp}
               moveStudent={moveStudent}
             />
           ))}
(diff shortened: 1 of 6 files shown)


frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -47,7 +47,15 @@ export const EditZing = () => {
       .get(`${API_ROOT}${COURSE_API}/students/${courseId}`)
       .then((res: AxiosResponse<CourseStudentDataResponse>) => {
         setUnmatchedStudents(res.data.data.unmatched)
-        setStudentGroups(res.data.data.groups)
+        let groups = res.data.data.groups.map((group: Group) => {
```
Coreball: Is there any particular reason for using `let` here?

frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -165,9 +173,15 @@ export const EditZing = () => {
...
         let newGroups = studentGroups.concat(response.data.data.groups)
-        console.log(response.data.data.unmatched)
+        newGroups = newGroups.map((group: Group) => {
+          return {
+            ...group,
+            shareMatchEmailTimestamp: group.shareMatchEmailTimestamp
+              ? new Date(group.shareMatchEmailTimestamp)
+              : null,
+          }
+        })
```
Coreball: I think newGroups might be a misnomer because in the line above it is actually `studentGroups.concat(response.data.data.groups)`. Coincidentally it still works because `new Date()` can take in an existing Date object, but we could make it clearer by only converting the new response data into Date objects.

Coreball (approved):
Looks very good! Getting the tooltips in as part of this PR is a nice addition too. Everything worked as I expected when testing, and using null to represent when an email hasn't been sent yet is a nice way to express it concisely. Thanks for getting this done so quickly.

frontend/src/modules/EditZing/Components/EditZing.tsx
mluo24: Nope, good catch

frontend/src/modules/EditZing/Components/EditZing.tsx
mluo24: Yeah I realize that now, I changed it so it only maps with the returned groups response.

PR #44, opened 2022-05-01
Edit page group selection checkboxes 

jewang25:
### Summary 

This pull request is the first step towards implementing feature to highlight group selected and make checkboxes appear. I added a list of selected groups that removes and adds groups based on whether the checkbox is selected or not. I still need to implement styling the group cards based on if the groups are selected or not and changing the button from email selected and email to.

### Test Plan

Test with multiple classes and multiple groups to check whether the groups being added to the selected groups list are correct or not. Checked the values in selected groups list in the console.
Screen Shot 2022-05-01 at 9 19 14 AM
Screen Shot 2022-05-01 at 9 19 20 AM

### Notes 
- will need to add styling based on whether the group is checked or not
- added a selected groups in edit zing and the checkbox in group grid

### DIFF
--- a/frontend/src/modules/EditZing/Components/EditZing.tsx
+++ b/frontend/src/modules/EditZing/Components/EditZing.tsx
@@ -38,16 +38,19 @@ export const EditZing = () => {
         setShowError(true)
       })
   }, [courseId])
+
   const [selectedGroups, setSelectedGroups] = useState<Group[]>([])
 
   const editSelectedGroups = (group: Group, selected: boolean) => {
     if (selected) {
       setSelectedGroups([...selectedGroups, group])
     } else {
-      setSelectedGroups(selectedGroups.filter((g) => g !== group))
+      setSelectedGroups(
+        selectedGroups.filter((g) => g.groupNumber !== group.groupNumber)
+      )
     }
   }
-  console.log(selectedGroups)
+  console.log(selectedGroups) //for seeing selected groups in console
 
   const [unmatchedStudents, setUnmatchedStudents] = useState<Student[]>([])
   const [studentGroups, setStudentGroups] = useState<Group[]>([])
@@ -250,7 +253,9 @@ export const EditZing = () => {
               moveStudent={moveStudent}
               createTime={studentGroup.createTime}
               updateTime={studentGroup.updateTime}
-              selected={selectedGroups.includes(studentGroup)}
+              selected={selectedGroups.some(
+                (g) => g.groupNumber == studentGroup.groupNumber
+              )}
               handleChecked={(event) => {
                 editSelectedGroups(studentGroup, event.target.checked)
               }}
--- a/frontend/src/modules/EditZing/Components/EditZing.tsx
+++ b/frontend/src/modules/EditZing/Components/EditZing.tsx
@@ -21,6 +21,7 @@ import {
 import { API_ROOT, COURSE_API, MATCHING_API } from '@core/Constants'
 import { useParams } from 'react-router-dom'
 import { MatchLoading } from './MatchLoading'
+import { Box, Button } from '@mui/material'
 
 export const EditZing = () => {
   const { courseId } = useParams<{ courseId: string }>()
@@ -233,10 +234,22 @@ export const EditZing = () => {
         courseNames={courseInfo.names}
         setShowMatchLoading={setShowMatchLoading}
       />
-      <StyledLogoWrapper>
-        <StyledLogo />
-        <StyledText>{courseInfo.names.join(', ')}</StyledText>
-      </StyledLogoWrapper>
+
+      <Box
+        display="flex"
+        flex-direction="row"
+        align-items=" center"
+        justifyContent="space-between"
+      >
+        <StyledLogoWrapper>
+          <StyledLogo />
+          <StyledText>{courseInfo.names.join(', ')}</StyledText>
+        </StyledLogoWrapper>
+        s
+        <Button sx={{ height: '40px', mt: '10px' }}>
+          {selectedGroups.length == 0 ? 'Send Email To' : 'Email Selected'}
+        </Button>
+      </Box>
       <DndProvider backend={HTML5Backend}>
         <Grid container spacing={1}>
           <UnmatchedGrid
--- a/frontend/src/modules/EditZing/Components/EditZing.tsx
+++ b/frontend/src/modules/EditZing/Components/EditZing.tsx
@@ -247,7 +247,7 @@ export const EditZing = () => {
         </StyledLogoWrapper>
         s
         <Button sx={{ height: '40px', mt: '10px' }}>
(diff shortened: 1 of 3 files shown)


gracshan:
Great work on this PR! I like your use of arrow functions and filtering to make the code concise and readable. This is a really useful feature, so I'm exited to see it in use!

frontend/src/modules/EditZing/Components/GroupGrid.tsx
```diff
@@ -60,8 +62,23 @@ export const GroupGrid = ({
...
+        style={{
+          opacity: isOver ? '0.6' : '1',
+        }}
       >
+        <Checkbox
+          sx={{
+            color: '#898992',
+            '&.Mui-checked': {
+              color: '#898992',
+            },
```
mluo24: You don't need this. If you set the "color" prop of the checkbox to be "secondary" (like `<Checkbox color="secondary" />`), you will get the correct color.

frontend/src/modules/EditZing/Components/GroupGrid.tsx
```diff
@@ -60,8 +62,23 @@ export const GroupGrid = ({
...
+            float: 'right',
+            width: '5 % ',
```
mluo24: I highly recommend you don't use "float" for styling in this case. What you want can be accomplished by using [flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/), which is a much better, more modern alternative.

frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -38,6 +38,16 @@ export const EditZing = () => {
         setShowError(true)
       })
   }, [courseId])
+  const [selectedGroups, setSelectedGroups] = useState<Group[]>([])
+
+  const editSelectedGroups = (group: Group, selected: boolean) => {
+    if (selected) {
+      setSelectedGroups([...selectedGroups, group])
+    } else {
+      setSelectedGroups(selectedGroups.filter((g) => g !== group))
```
mluo24: Having this as a check makes this sort of iffy. See my review above for more details. Also see the "include" part when you pass down the selected prop.

mluo24 (changes requested):
Good job on this feature, as emails are super important right now! I found a bug where if you edit a group (drag a member to another group), it will uncheck itself, but still leave itself in the array, causing a bunch of garbage data (groups that don't exist anymore) to accumulate. I think this might be an issue of asking the designers whether a user should be able to edit groups while they have them checked or not, but I feel like there wouldn't be a reason to not allow it. To fix this, I recommend searching by group number as opposed to naively checking if the whole group is equal. That way, the group number is guaranteed to be the same, and your implementation would be invariant to changes made to the group.

As for the styling, I would recommend you check out [this page](https://mui.com/material-ui/customization/how-to-customize/#dynamic-overrides) for inspiration, as you can use a "checked" prop to style the actual group box when the checkbox is checked. I think there's also a &:hover pseudoselector you can use for MUI for hovering.

frontend/src/modules/EditZing/Components/EditZing.tsx
Coreball: For the "include" part, [`.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) might be useful

Coreball (commented):
Good work! Thanks to @mluo24 for finding that bug as well, this is why we have pull request reviews because that didn't occur to me when I was initially thinking about the implementation. I agree that using group number is a good way to fix it. This is also why when reviewing code the reviewers should test the changes locally, or at least try the preview on Firebase Hosting. Miranda has good points about styling too. You have done great work on this PR! Remember that we are critiquing the code, you have done an amazing job.

frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -220,10 +234,22 @@ export const EditZing = () => {
...
+        display="flex"
+        flex-direction="row"
+        align-items=" center"
+        justifyContent="space-between"
+      >
+        <StyledLogoWrapper>
+          <StyledLogo />
+          <StyledText>{courseInfo.names.join(', ')}</StyledText>
+        </StyledLogoWrapper>
+        s
```
mluo24, line 248: Remove this "s" lol

frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -39,6 +40,19 @@ export const EditZing = () => {
...
+  const editSelectedGroups = (group: Group, selected: boolean) => {
+    if (selected) {
+      setSelectedGroups([...selectedGroups, group])
+    } else {
+      setSelectedGroups(
+        selectedGroups.filter((g) => g.groupNumber !== group.groupNumber)
+      )
+    }
+  }
+  console.log(selectedGroups) //for seeing selected groups in console
```
mluo24, line 54: You can remove the console.log once you're done with everything (do it before you merge it in for sure).

frontend/src/modules/EditZing/Components/GroupGrid.tsx
```diff
@@ -79,6 +103,18 @@ export const GroupGrid = ({
...
+            color="secondary"
+            align-self="flex-end"
+            checked={selected}
+            onChange={handleChecked}
+            sx={{
+              display: selected || isHovering ? 'block' : 'none',
+              padding: '0px',
+              ml: '45%',
+            }}
+          />
```
mluo24, line 117: One neat trick I really like using Flexbox with placing two elements on the left and right ends of each other is adding `<Box flexGrow={2} />` in between the two elements you want to separate (so in this case the Checkbox and title components, and I guess the tooltips too) which will essentially fill up the space in between, forcing the two elements separated by it to the opposite ends. With that, you can remove `align-self`.

mluo24 (approved):
Really good job on this pull request! I love how you were able to integrate Changyuan and my suggestions to create an elegant implementation! It works pretty smoothly for the most part, and I left some general comments, mostly about code cleanup and using MUI. 

Unfortunately I don't think all of the bugs have been squashed necessarily yet, as when I do try and edit groups, the groups (as seen in the selectedGroup state) are not updated with the changes. Since the only time a group is edited (meaning it exists already and a change occurred to it) is with the drag and drop, so I think the easiest fix would be to update the selected groups every time a drag and drop happens. This can likely be done by passing down a function to the GroupGrid component and using it in the `useDrop` hook by line 51.

I'm approving it right now since it's a bug that's more of an edge case and not necessarily a big deal (since you do have the same number of groups that you checked still, just the wrong people are in it), but if you're up to fixing it, go for it! Again, amazing work!

azhou35 (commented):
Looks great Jessica! This feature will be important for the email feature that is crucial to our summer launch. Once you move onto more detailed styling it'll be great to consult the designers and their Figma designs, if you haven't already. Keep up the good work 💪🏻

PR #46, opened 2022-05-01
Initial implementation of filter matching

jessica0801:
### Summary 

This pull request is the  first step towards implementing filtering for the class cards on the zing-lsc dashboard. It allows users to filter the  class cards available by alphabetical, reverse alphabetical, unmatchable first, newly matchable first, and matchable first. The feature for filtering by newest has still yet to be implemented. 

The filtering in this pull request is completed through customizing the   localeCompare() function and using it to sort a copied version of the groups array. The  matchability status is implemented such that a card will be sorted first if it has the   desired qualities. 

Screen Shot 2022-05-01 at 10 07 56 AM
Screen Shot 2022-05-01 at 10 08 08 AM

Screen Shot 2022-05-01 at 10 10 55 AM

This pull request is the first step towards implementing filter by matchability status

- [x] implemented filtering alphabetically
- [x] implemented filtering reverse alphabetically
- [x] implemented unmatchable first
- [x] implemented newly matchable first
- [x] implemented matchable first

### Test Plan 

Added several iterations of each category to see if the   sorting resulted in the  right order depending on the filtering status.

### Breaking Changes

### DIFF
--- a/frontend/src/modules/Dashboard/Components/Dashboard.tsx
+++ b/frontend/src/modules/Dashboard/Components/Dashboard.tsx
@@ -43,43 +43,83 @@ export const Dashboard = () => {
   }, [])
   // (a,b) = -1 if a before b, 1 if a after b, 0 if equal
   function sorted(groups: CourseInfo[], menuValue: string) {
-    if (menuValue === 'newly-match-first') {
-      return [...groups].sort((a, _) => {
-        //-1 if a newly matchable and b isn't
-        if (a.lastGroupNumber === 0 && a.unmatched.length > 1) {
-          return -1
-        } else return 1
-      })
-    } else if (menuValue === 'unmatch-first') {
-      return [...groups].sort((a, _) => {
-        //-1 if a unmatchable and b isn't
-        if (a.lastGroupNumber === 0 && a.unmatched.length === 1) {
-          return -1
-        } else return 1
-      })
-    } else if (menuValue === 'match-first') {
-      return [...groups].sort((a, _) => {
-        //-1 if a matchable and b isn't
-        if (a.lastGroupNumber > 0 && a.unmatched.length > 0) {
-          return -1
-        } else return 1
-      })
-    } else if (menuValue === 'classesa-z') {
-      const sorted = [...groups].sort((a, b) => {
-        return a.names[0].localeCompare(b.names[0], undefined, {
-          numeric: true,
+    switch (menuValue) {
+      case 'newly-match-first':
+        return [...groups].sort((a, _) => {
+          //-1 if a newly matchable and b isn't
+          if (a.lastGroupNumber === 0 && a.unmatched.length > 1) {
+            return -1
+          } else return 1
         })
-      })
-      return sorted
-    } else if (menuValue === 'classesz-a') {
-      const sorted = [...groups].sort((a, b) => {
-        return b.names[0].localeCompare(a.names[0], undefined, {
-          numeric: true,
+      case 'unmatch-first':
+        return [...groups].sort((a, _) => {
+          //-1 if a unmatchable and b isn't
+          if (a.lastGroupNumber === 0 && a.unmatched.length === 1) {
+            return -1
+          } else return 1
         })
-      })
-      return sorted
-    } else return groups
+      case 'match-first':
+        return [...groups].sort((a, _) => {
+          //-1 if a matchable and b isn't
+          if (a.lastGroupNumber > 0 && a.unmatched.length > 0) {
+            return -1
+          } else return 1
+        })
+      case 'classesa-z':
+        return [...groups].sort((a, b) => {
+          return a.names[0].localeCompare(b.names[0], undefined, {
+            numeric: true,
+          })
+        })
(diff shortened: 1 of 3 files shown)


jewang25 (commented):
The filter function looks good! I like how it can filter based on match ability status now. I also like how you added comments to your code to make it clear what it does in certain situations.

wizhaaa (commented):
Awesome pr, great use of mui components, + making the code readable. Really helpful pr to move forward on the dashboard functionality! I think the only thing you could change (not necessary tho) is to change the ``else ifs`` into either a switch or dispatch function but don't think the performance difference is that big either way.

CollinWoo:
Great work! The filtering works perfectly for me and the code is very clear and readable. But I agree with Will's suggestion to replace the `else if` statements with a switch statement since it reduces redundancy and makes extending the filter functionality easier, especially with the long chain of `menuValue` strict equality comparisons in Dashboard.tsx. Other than that, everything looks great!

frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -38,11 +41,77 @@ export const Dashboard = () => {
       setGroups(res.data.data)
     })
   }, [])
+  // (a,b) = -1 if a before b, 1 if a after b, 0 if equal
+  function sorted(groups: CourseInfo[], menuValue: string) {
+    if (menuValue === 'newly-match-first') {
```
b3njamint: Perhaps you could turn this stuff into a series of switch statements to make it more readable?

b3njamint (commented):
Good job Jessica! I know it will definitely play a big role in the future when we have more classes inputted. The code is very readable and I appreciate the simplicity + comments.

Edit:  Sorry I didn't see the previous comments—it seems that others already pointed out the switch statement stuff. Please ignore.

frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -38,11 +41,77 @@ export const Dashboard = () => {
...
+    setGroups(sorted(groups, event.target.value))
+  }
 
   return (
     <StyledContainer>
       <StyledHeaderMenu>
         <LogoImg />
+        <Box sx={{ display: 'flex', flexDirection: 'row' }}>
+          <Box
+            sx={{ fontWeight: 'bold', color: '#5C5B6A', padding: 1, margin: 1 }}
```
Coreball: @mluo24 knows more than I do, but I believe you can reference this color with `color: 'essentials.75'` instead of hard coding it, since `sx` has access to the theme

frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -16,8 +16,11 @@ import { KeyboardArrowDown } from '@mui/icons-material'
 import { logOut } from '@fire'
 import { useAuthValue } from '@auth'
 import { useHistory } from 'react-router'
+import { DropdownSelect } from '@core/Components'
+import { SelectChangeEvent, Box } from '@mui/material'
 
 export const Dashboard = () => {
+  const [status, setStatus] = useState('')
```
Coreball: When I first read this I wasn't sure what `status` meant, maybe a variable name like `sortOrder` might be better?

frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -38,11 +41,77 @@ export const Dashboard = () => {
...
+        return b.names[0].localeCompare(a.names[0], undefined, {
+          numeric: true,
+        })
+      })
+      return sorted
+    } else return groups
+  }
+  const handleChange = (event: SelectChangeEvent) => {
+    setStatus(event.target.value)
+    setGroups(sorted(groups, event.target.value))
```
Coreball: Thought of a better way to do the sorting and not modify the original group array:
```js
const handleChange = (event: SelectChangeEvent) => {
  setStatus(event.target.value)
  // setGroups(sorted(groups, event.target.value))
}

const sortedGroups = sorted(groups, status)
```
Here the dropdown only changes the sort order, then every re-render a copy of the groups are sorted and used as `sortedGroups`, which is used for the groups component later. This makes the sorting possibly more predictable when there are ties because the previous sort of `groups` doesn't influence the next sort of `groups`, since `setGroups` is only called when loading the data.

Coreball (commented):
Good job! Filtering/sorting by matchability status is going to be really important, especially when there are dozens of classes students may be requesting study partners for. I agree with the other reviewers that switching to a switch (hah) might make the code easier to read. I know last time we talked we decided to handle the groups state in this way, but I thought of a more preferable way to do it that preserves one-way data flow even better (see specific comment).

Also, is this supposed to happen? I was a bit confused because I thought CS 2110/CS 1110 in this example would be listed ahead of the class that has only one student. The sort for unmatchable and matchable first produces the same sort order in this situation.
Screen Shot 2022-05-04 at 16 04 28

Coreball (approved):
Looks good! I think the code is better to read now.

PR #48, opened 2022-05-03
Removed duplicates from cross-listed classes

b3njamint:
### Summary 

This pull request fixes the duplication issues presented when a student inputs multiple classes that are cross-listed. For example, when a student inputs "CS 2110" and "ENGRD 2110", they will be shown to belong to two groups.

I spent a lot of time trying to figure out where the issue was and how to fix it. My solution 

- [x] fixed duplication error from cross-listed classes

### Test Plan 

Originally, this happens (cross liste coutses: CS 2110, ENGRD 2110)
Screen Shot 2022-05-03 at 6 45 08 PM

Now, duplicates are gone! (cross listed courses: ARCH 3702, CS 1620, ART 2907, ENGRI 1620)
Screen Shot 2022-05-03 at 6 46 11 PM

I also tested resubmitting the form and added duplicate classes. Nothing new was added, as expected.

### Notes 

I added some comments throughout the two files I touched so future developers would know what is going on. There are a lot of variables.

### Breaking Changes 

Don't think there are any breaking changes.


backend/functions/src/student/functions.ts
```diff
@@ -37,27 +37,32 @@ const addStudentSurveyResponse = async (
   studentRef = studentR
 ) => {
   // First, update the [student] collection to include the data for the new student
+  // Converts to set first to eliminate duplicates, then converts to list
```
Coreball: Is this comment still accurate on the latest version of the code?

Coreball (commented):
Good job Benji! Before I approve this, I'd like to point out that this code caused an error in the function logs when I tried ['INFO 2040', 'CS 2850'], however I don't think it's specific to those courses. I looked into it a bit and I believe it's caused by line 93, which is `.doc(crsesToAdd[index].toString())`. I think it is related to array indices and a mismatch between `crseCatalogsToUpdate` and `crsesToAdd`, since they appear to be parallel lists. Other than that, I especially like how you iterated through various methods of solving this bug before settling on the last one, which is simple and easy to understand with not many additional lines!

Insomnia
Screen Shot 2022-05-03 at 23 15 10

Function log output
Screen Shot 2022-05-03 at 23 15 42

wizhaaa (approved):
gud fix

hexbabe (approved):
LGTM I tried testing it locally with https://classes.cornell.edu/browse/roster/SP22/class/PSYCH/1102
Screen Shot 2022-05-24 at 7 18 09 PM

PR #49, opened 2022-05-23
Change default roster FA21 -> SP22

Coreball:
### Summary 

FA21 roster broke so quick fix is to change it to SP22 for dev purposes. Before launch will have to update it again.

- [x] default roster changed to `SP22`

### Test Plan 

My emulator is not working so no tests lol


hexbabe (approved):
This is necessary bc it seems that the FA21 API is down and the students endpoints r borken! bork bork

PR #50, opened 2022-05-24
micro soft log in 

wizhaaa:
### Summary 

Changes login method to microsoft in order to obtain auth code to send emails. 

This pull request is the first step the sending emailing feature. 

- [x] Microsoft login
- [x] removed google login 

### Test Plan 

Delete users in firebase -> login again -> redirects to dashboard successfully. 

Logged in? Log out and then keep going. 

Testing also had console logs in each auth checker as well to check if passed or not. 

Also tried in incognito mode. 

### Notes 

Our auth settings in firebase have one account per email so you must delete your google account in the Firebase users in order to complete ms login.


Coreball (approved):
Looks good. I think changing the text to say "Staff Login" is a good choice as well, in case students accidentally end up on the home page.

PR #52, opened 2022-05-25
Group checkboxes

Coreball:
### Summary 

This pull request fixes a bug remaining in #44 and works on styling the group cards and the edit page.

- [x] fixed drag and drop group selection bug
- [x] updated group card styling (css transitions are cool)
- [x] updated edit page styling

### Test Plan 

To test the drag and drop bug being fixed, I logged the selected group content before and after performing the student drag and drop movement. The selected groups updated to reflect the actual group membership.

Styling Before

https://user-images.githubusercontent.com/22627336/170176478-cce3b337-ea35-4faf-9005-ffeedb41be55.mov

Styling After

https://user-images.githubusercontent.com/22627336/170176486-31c40aaa-f96d-4f3e-ab1e-b0d76699fdf5.mov

PR #53, opened 2022-05-25
Dropdown email

Coreball:
### Summary 

This pull request adds the menu drop down that says "Send email to" and "Email selected" on the edit page. At the moment there is only one option that says "Newly Matched". For now, the onClick of the "Email selected" only logs a message to the console. After #52 is merged in, the diff should be clearer.

- [x] implemented menu drop down send email to
- [x] updated Menu styling theme

### Test Plan 

https://user-images.githubusercontent.com/22627336/170385809-d0e96514-f793-4529-b344-178deed605a9.mov

### DIFF
--- a/frontend/src/modules/Core/Components/DropdownSelect.tsx
+++ b/frontend/src/modules/Core/Components/DropdownSelect.tsx
@@ -42,13 +42,17 @@ export const DropdownSelect = ({
           MenuListProps: {
             dense: true,
             sx: {
-              borderRadius: '10px',
               '& .MuiMenuItem-root': {
                 fontFamily: 'Montserrat',
                 '&:hover': {
                   bgcolor: 'purple.16',
                 },
               },
+            },
+          },
+          PaperProps: {
+            sx: {
+              borderRadius: '10px',
             },
           },
         }}
--- a/frontend/src/modules/EditZing/Components/EditZing.tsx
+++ b/frontend/src/modules/EditZing/Components/EditZing.tsx
@@ -18,11 +18,14 @@ import {
   Box,
   Button,
   Grid,
+  Menu,
+  MenuItem,
   SvgIcon,
   SvgIconProps,
   Typography,
 } from '@mui/material'
 import { ReactComponent as Lsc } from '@assets/img/lscicon.svg'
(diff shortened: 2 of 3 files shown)


frontend/src/modules/Core/Components/DropdownSelect.tsx
```diff
@@ -51,6 +50,11 @@ export const DropdownSelect = ({
               },
             },
           },
+          PaperProps: {
+            sx: {
+              borderRadius: '10px',
+            },
+          },
```
mluo24, line 57: Was there any reason for adding this here?

frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -225,8 +232,32 @@ export const EditZing = () => {
...
+  }
+
+  const handleSelectNewlyMatched = () => {
+    setSelectedGroupNumbers(
+      studentGroups
+        .filter((group) => group.shareMatchEmailTimestamp === null)
+        .map((group) => group.groupNumber)
+    )
+    handleMenuClose()
+  }
```
mluo24, line 258: Is the definition of newly matched just "which one did not have an email get sent to"? I'm just wondering since I had no idea what the basis for the criterion is.

mluo24 (approved):
Nice work implementing this, Changyuan, especially with wrestling with the MUI component styling in theme overrides. I think you branched this off of the group checkboxes, so make sure if you make any changes in that branch again and merge that they merge properly.

frontend/src/modules/Core/Components/DropdownSelect.tsx
Coreball: The previous sx borderRadius for MenuListProps wasn't actually changing what the drop down looked like as it was supposed to

frontend/src/modules/EditZing/Components/EditZing.tsx
Coreball: My understanding is yes, that groups are considered newly matched if they haven't been sent an email telling them that they are matched with each other yet.

Coreball:
Rebased to `main` because I wanted that clean diff

PR #54, opened 2022-05-27
Add script to populate Firebase emulator with course + student data

mluo24:
### Summary 

This pull request fixes the previously made `add-test-data.js` file to be written in TypeScript as well as adding an npm run script to run the file so that it could populate the database. Type `npm run populate` for this script to run and see the changes in the Firebase emulator. The rational is to get quick data populated so you don't need to enter a bunch of survey responses to get a group.

### Notes 

From https://firebase.google.com/docs/emulator-suite/connect_firestore#node.js-admin-sdk

This sets a variable, not sure if this affects anything.

### DIFF
--- a/backend/functions/package.json
+++ b/backend/functions/package.json
@@ -14,7 +14,8 @@
     "start": "npm run shell",
     "deploy": "firebase deploy --only functions",
     "logs": "firebase functions:log",
-    "prepare": "cd ../.. && husky install backend/functions/.husky"
+    "prepare": "cd ../.. && husky install backend/functions/.husky",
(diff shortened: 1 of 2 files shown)


Coreball (approved):
Looks good, thank you so much for doing this! It will make manual testing so much easier. One thing I would like to point out though is that the survey currently handles class years like the following (because of grad students?)

```json
"answers": {
  "Freshman": "1st year undergraduate",
  "Sophomore": "2nd year undergraduate",
  "Junior": "3rd year undergraduate",
  "Senior": "4th year undergraduate",
  "Grad": "Grad student"
}
```

PR #56, opened 2022-05-30
MS Send Mail Email button 

wizhaaa:
### Summary 

PR Current Structure: 

Created a mock front-end page ("/email") to display the send email button. 
Created a backend endpoint (POST at "/sendmail") that the frontend must send email data to. 
This button calls the "send mail" function in the frontend which posts to our backend server to send the email through Graph API. 

Backend will then respond with success or failure. 
How the front end will decide to render based on the response is left to a future PR. 
One option I have implemented is to render a "Try Again" button that will log the user in (repopulating authToken) and automatically call the sendMail function again with the same data. 

Next: 
Once this and Sean's PR #57 are merged, will move the sendMail function to be accessible and obtain the recipients, body, and subject in Sean's code. 

**TODO:** What data the front end needs to pass to this function and update: (Will implement and obtain once #57 is merged). 

- [ ] email subject
- [ ] email body
- [ ] email rcpts of the students 

Right now, the front end passes in static dummy text and sends to my emails. 

Note: Only I have access to the lscstudypartners mailbox so I believe nobody else on Zing will be able to send emails successfully. Would want to allow other Zing devs to be owners of mailbox too and help in testing? Or, just hand off to LSC for testing with their MS accounts. 

ALSO NOTE: auth account exists with different credential error: 
We have enabled one account per email so to login with MS, you must delete your google account in the authentication -> users in firebase.

### Test Plan 

User with invalid auth token: 
Screen Shot 2022-05-30 at 2 52 36 AM

Valid auth token and email sent succ: 
Screen Shot 2022-05-30 at 3 03 18 AM

Email showing as sent from LSC: 
Screen Shot 2022-05-30 at 3 04 05 AM

### TODO 
To be able to use this for and incorporate into our app: need to either pass email body, subject, and recipient emails (string list) into this email button function/compoenent or whatever and replace the static dummy/fake data I have now. 

Also, to decide how many states we will need & how to handle email sending err/success. 

Video: https://cornelldti.slack.com/files/U02KJGFHUNT/F03HK6R0KFD/screen_recording_2022-05-30_at_6.01.14_pm.mov

https://user-images.githubusercontent.com/46132945/171063372-7053db55-654a-49ef-b020-f01d27d6840a.mov


backend/functions/src/emailing/functions.ts
```diff
@@ -0,0 +1,89 @@
...
+/* Create Recipients parses the rcpts (string list of student emails) 
+    sent from the frontend. Called in addRecipients. */
+function createRecipients(rcpts: string[]) {
+  return rcpts.map((rcpt) => {
+    return {
+      emailAddress: {
+        address: rcpt,
+      },
+    }
+  })
```
Coreball: Personally I like omitting the return statement and mapping to a new value directly in cases like these:
```js
.map((rcpt) => ({ emailAddress: { address: rcpt } })
```
Your way is fine tho

frontend/src/firebase/firebase.js
```diff
@@ -18,25 +18,23 @@ export async function signInWithGoogle() {
 }
 
 export async function signInWithMicrosoft() {
+  let accessToken
```
Coreball: Do you need this line if accessToken is being assigned once

frontend/src/firebase/firebase.js
```diff
@@ -18,25 +18,23 @@ export async function signInWithGoogle() {
...
-      console.log('access token: ' + accessToken)
-      console.log('id token: ' + idToken)
-    })
-    .catch((error) => {
-      // handle error
-      console.log('error in ms login' + error)
-    })
+export async function adminSignIn() {
+  await signInWithMicrosoft().then((res) => {
+    window.localStorage.setItem('authToken', res)
```
Coreball, line 34: Sometimes calling it accessToken and sometimes calling it authToken could be confusing in the future. Is there a reason for the two names or can it be standardized?

backend/functions/src/emailing/functions.ts
```diff
@@ -0,0 +1,89 @@
...
+      logged in user
+    
+    Sends a POST request to the GRAPH API. 
+    
+    Returns: 
+      202 (:number) if successfully sent. 
+      Other if email failed to send. 
+        - Bad AUTH or else. Frontend will then parse this response 
+        - Either relogin once (?) 
+      */
```
Coreball, line 61: Documentation is nice

frontend/src/modules/Emailing/Components/Emailing.tsx
```diff
@@ -0,0 +1,207 @@
...
+    FAIL -> 401  */
+export const sendEmail = async (
+  emailSubject: string,
+  emailRcpts: string[],
+  emailBody: any,
+  emailSent: boolean,
+  setEmailSent: (arg: boolean) => void,
+  retried: boolean,
+  setRetried: (arg: boolean) => void
+) => {
```
Coreball: Yeah I think to be more generalizable this function could just focus on the part where it gets the token from local storage. Might want to leave the user email stuff to a function input so it can be obtained in the same way that it's obtained elsewhere in the app

Coreball (approved):
Nice work. Important building blocks for emailing and I like the demo page, especially the code comments that detail what's happening in each step. As you noted I was not able to test it myself, but your video was very convincing. Gj, I like to see it actually working.

PR #57, opened 2022-05-30
Email modal

hexbabe:
### Summary 

This pull request is the first step towards implementing the email modal + template selector sending user flow. All basic functionality for the summer launch is done and hopefully bug-less

Run `yarn` again for good measure

Tried to use component system ZingModal, but the functionality was limited so I had to reach in and change some of its props and modularity.

Further, the feeding the buttons state props causes re-rendering of the entire modal when those states change, so I had to move the buttons into the child component instead.

This all being said, it works now. Most of my changes were in the EditZing dir with new "Email-" prefixed files and styles. I tried to adhere to the new MUI way of styling (with some named styledcomponents here and there) but lmk if things are not implemented correctly with this new system

- [x] implemented basic functionality
- [x] connect with mailer + new edit page (I commented the entrypoint + exitpoints on this PR below)
- [ ] roll out the rest of the templates eventually
- [ ] implement custom message
- [ ] implement editing master template

### Test Plan 

Use insomnia `addStudentSurveyResponse` endpoint to populate class with a ton of groups (5 or more). Make sure that selectedGroups is passed along correctly to the modal and the modal styling looks correct for Step 0. After selecting the template, the next button should become active. Now, on step 1, we should see that the template name is correct and the email preview is also correct (in respect to the template name). Pressing "Send Emails" should then close the modal. Pressing "Back to templates" should take you back to Step 0 and all functionality of Step 0 should be preserved.

Video demo: https://www.loom.com/share/f5f93634ee8c457483b199ece521df6e
Figma node: https://www.figma.com/file/9zueLCvKT7BBMYi8Dt8Ukp/Zing-LSC?node-id=3073%3A227423


frontend/src/modules/Core/Components/ZingModal.tsx
```diff
@@ -13,30 +13,27 @@ import CloseIcon from '@mui/icons-material/Close'
 import React from 'react'
 import { ZingModalProps } from '@core'
 
-const TempDialog = styled(Dialog)(({ theme }) => ({
```
hexbabe, line 16: moved this into the component in order to access props for dynamic styling

frontend/src/modules/Core/Components/ZingModal.tsx
```diff
@@ -58,20 +55,47 @@ const TempDialogTitle = (props: DialogTitleProps) => {
...
+  }))
+
   return (
     <TempDialog
       onClose={props.onClose}
       open={props.open}
       BackdropProps={{
         sx: {
-          backgroundColor: 'rgba(0, 0, 0, 0.7)',
+          backgroundColor: 'rgba(0, 0, 0, 0.3)',
```
hexbabe: adheres to figma better

frontend/src/modules/Core/Components/ZingModal.tsx
```diff
@@ -58,20 +55,47 @@ const TempDialogTitle = (props: DialogTitleProps) => {
...
           '-webkit-tap-highlight-color': 'transparent',
         },
       }}
-      sx={props.sx}
+      // imo sx should go to DialogContent rather than the outermost container
```
hexbabe: sx prop*

frontend/src/modules/EditZing/Components/EmailModalContent.tsx
```diff
@@ -0,0 +1,151 @@
...
+}: EmailModalContentProps) => {
+  const [selectedTemplate, setSelectedTemplate] = useState<TemplateName>(
+    '' as TemplateName
+  )
+  const [step, setStep] = useState<number>(0)
+  const titles = ['Select email template', 'Send emails']
+
+  const RecipientsComponent = () => {
+    let groupNames = ''
+    selectedGroups.forEach((group, index) => {
```
hexbabe: tried to use .reduce here but unfortunately the return type is not of type Group, so :/

frontend/src/modules/EditZing/Components/EmailModalContent.tsx
```diff
@@ -0,0 +1,153 @@
...
+        >
+          Next
+        </Button>
+      )
+    } else {
+      return (
+        <Button
+          onClick={() => {
+            setIsEmailing(!isEmailing)
+            // add something else here that actually sends the emails and a pop-up message saying that it's in-progress then done/fail etc.
```
hexbabe: this would be the exit-point that connects to @notwz mailer

frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -227,6 +229,14 @@ export const EditZing = () => {
 
   return courseInfo && hasLoadedStudentData ? (
     <StyledContainer>
+      {isEmailing ? (
```
hexbabe: Entrypoint 1 for connecting to new button + edit

frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -245,8 +255,10 @@ export const EditZing = () => {
           <StyledLogo />
           <StyledText>{courseInfo.names.join(', ')}</StyledText>
         </StyledLogoWrapper>
-        s
-        <Button sx={{ height: '40px', mt: '10px' }}>
+        <Button
+          sx={{ height: '40px', mt: '10px' }}
+          onClick={() => setIsEmailing(!isEmailing)}
```
hexbabe: Entrypoint 2 for connecting to new button + edit

wizhaaa (approved):
Email modal functionality is working so far for me. I suppose I just have to parse the selectedGroups to get each students email and use the email utils to get the email body. Can instantiate that data in the EmailModalContent -> import my emailing function -> pass that data in -> and return an error or success message.

wizhaaa:
also one comment: when selecting groups, the check box only shows up when you hover the group card. may not be intuitive ? unless that is the designers choice. idk just wanted to lyk in case. 

also right now you can send emails when no groups are selected, maybe block it/show a little error msg on the bottom when they try but not a big deal/ doesn't break the functionality so no big deal except maybe that there are no recipients if they want to send an email 

also edit zing student cards are facing the the text clipping issue but may be for future pr ?

Coreball:
@notwz the checkbox only showing up when hovered or selected is intentional design choice. In #53 the send email to button only shows up when at least one group is selected (no support for individual students yet) so once this branch gets those updates from main it should be all good.

frontend/src/modules/Core/Components/ZingModal.tsx
Coreball: According to styled components you should not do this, because it re-creates the TempDialog every render pass. https://styled-components.com/docs/basics#define-styled-components-outside-of-the-render-method (and the example is for a regular styled.div too, not even a more complicated Dialog component)

```
// WARNING: THIS IS VERY VERY BAD AND SLOW, DO NOT DO THIS!!!
```

frontend/src/modules/Core/Components/ZingModal.tsx
Coreball: @mluo24 what do you think

frontend/src/modules/EditZing/Components/EditZing.tsx
Coreball: Instead of
```js
{isEmailing ? <...> : null}
```
You could use
```js
{isEmailing && <...>}
```

frontend/src/modules/EditZing/Components/EditZing.tsx
Coreball: Merge from main into ur branch and update with the new code here

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -0,0 +1,29 @@
...
+      containerWidth={'800px'}
+      containerHeight={'580px'}
+      children={
+        <EmailModalContent
+          selectedGroups={selectedGroups}
+          courseNames={courseNames}
+          isEmailing={isEmailing}
+          setIsEmailing={setIsEmailing}
+        />
+      }
```
Coreball: I assume you know that the children prop means
```js
<ZingModal children={<p>Zing</p>} />
```
Is equivalent to
```js
<ZingModal>
  <p>Zing</p>
</ZingModal>
```
And you wrote it this way for a reason, right

frontend/src/modules/EditZing/Components/EmailModalContent.tsx
Coreball:
```js
const groupNames = selectedGroups
  .map(group => `Group ${group.groupNumber}`)
  .join(', ')
```

frontend/src/modules/EditZing/Components/EmailModalContent.tsx
```diff
@@ -0,0 +1,153 @@
...
+        <Typography variant="h5" component="h5" fontWeight={'900'}>
+          {groupNames}
+        </Typography>
+      </InfoContainer>
+    )
+  }
+
+  const TemplateSelectedComponent = () => {
+    return (
+      <InfoContainer style={{ marginTop: '0.25em', marginBottom: '0.5em' }}>
```
Coreball: Don't feel like InfoContainer/RecipientContainer really needs to be its own reusable component if one of the two places where it's reused, it has style overrides anyway (plus it's just a `display: flex`? Use a Box?

frontend/src/modules/EditZing/Components/EmailModalContent.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+export const EmailModalContent = ({
+  selectedGroups,
+  courseNames,
+  isEmailing,
+  setIsEmailing,
+}: EmailModalContentProps) => {
+  const [selectedTemplate, setSelectedTemplate] = useState<TemplateName>(
+    '' as TemplateName
+  )
```
Coreball: I feel like the `'' as TemplateName` is abusive of TypeScript because we are basically convincing TypeScript that whatever's on the left hand side will be of the correct type, except here we already know `''` is not part of the enum type.

frontend/src/modules/EditZing/Components/EmailPreview.tsx
```diff
@@ -0,0 +1,57 @@
...
+    borderRadius: '5px 5px 0px 0px;',
+  }
+  const BodyAndSubjectSx: SxProps = {
+    color: 'essentials.75',
+    backgroundColor: 'essentials.6',
+    border: '0.5px solid',
+    borderBottom: '0.5px solid #898992;',
+    borderColor: 'essentials.50',
+    padding: '16px',
+  }
```
Coreball, line 28: Alternatively you could pass it a function `theme => ({ ... })` instead of just the object if you want to take advantage of things like `padding: 2` which multiplies the theme's default spacing by 2 (default spacing is 8). Don't need to do this though

frontend/src/modules/EditZing/Components/EmailPreview.tsx
```diff
@@ -0,0 +1,57 @@
...
+      >
+        <Box display={'flex'} flexDirection={'row'}>
+          <Typography fontWeight={800}>Subject: </Typography>&nbsp;
+          <Typography>Study Partners!</Typography>
+        </Box>
+      </Box>
+      <Box
+        sx={{
+          maxHeight: '250px',
+          overflow: 'scroll',
```
Coreball, line 48: I like the overflow here, but just setting `overflowY` looks nicer IMO

frontend/src/modules/EditZing/Components/EmailTemplateButtons.tsx
```diff
@@ -0,0 +1,40 @@
...
+      }}
+    >
+      <Typography variant="h5" component="h5" fontWeight="800" mb="12px">
+        Use an existing template:
+      </Typography>
+      <RadioButtons
+        values={activeTemplates}
+        onClick={handleChange}
+        key={'TemplateRadioButtons'}
+        currentAnswer={selectedTemplate || ''}
```
Coreball: What does this `selectedTemplate || ''` do if `selectedTemplate` is of type string and falsy strings are already `''`

frontend/src/modules/EditZing/utils/emailTemplates.tsx
```diff
@@ -0,0 +1,252 @@
...
+      reply to this email to let us know what’s going on, and we’ll try to help.
+    </div>
+      <br> 
+      <br> 
+    <div> 
+      Take care, 
+      The LSC Staff
+    </div>
+    `
+}
```
Coreball: Does this file have to be named `.tsx`? As far as I can tell there's no JSX in the code and it breaks GitHub's syntax highlighting because the first character of the filename is not capitalized like a react component file

frontend/src/modules/EditZing/utils/emailTemplates.tsx
```diff
@@ -0,0 +1,252 @@
...
+  MATCHED = 'Share matched results',
+  FIRST_NO_MATCH = 'First no match notification',
+  SECOND_NO_MATCH = 'Second no match notification',
+  ADD_STUDENT = 'Request to add student to group',
+  LATE_ADD_STUDENT = 'Request to add student to group (late)',
+  ASK_JOIN_GROUP = 'Ask to join group',
+  CHECK_IN = 'Check in with groups',
+}
+
+export const getBody = (templateName: string, className: string) => {
```
Coreball: Should `templateName` be of type `TemplateName`? I also wonder if the enum could also be a union type of string literals

Coreball (approved):
Approving this for now because it works and we want to move somewhat quickly at the start of the summer. Overall good job, definitely made those designs a reality. Sometimes I know how to easily do assigned developer tasks and sometimes not, and this is the second case, so nice work. Most of my comments are relatively minor, but I would definitely like to see that use of styled() moved out of the function component body.

Also wish the correct GitHub Actions would run

frontend/src/modules/Core/Components/ZingModal.tsx
mluo24: This shouldn't matter anymore. I think the only thing that needs to be amended is the height and width of the container, which is already taken care of by the new props. Everything else can use the new components.

mluo24 (commented):
Take a look at https://github.com/cornell-dti/zing-lsc/pull/59. You're free to merge it/modify as you'd like.

PR #58, opened 2022-06-01
Finish updating student survey for launch

CollinWoo:
### Summary 

This pull request implements the survey design shown [here](https://www.figma.com/file/9zueLCvKT7BBMYi8Dt8Ukp/Zing-LSC?node-id=2394%3A234446). See http://localhost:3000/survey.

- [x] Disable next button if no option is selected
- [x] Make it look like designs (mostly complete, see notes)
- [x] Make submission confirmation screen an actual confirmation that could fail if the request doesn’t succeed

### Test Plan 

Changes were mostly visual so testing was done by comparing the Figma to the actual frontend. Screenshots of the frontend implementation are below:

Screen Shot 2022-05-31 at 10 19 24 PM
Screen Shot 2022-05-31 at 10 19 36 PM
Screen Shot 2022-05-31 at 10 19 48 PM
Screen Shot 2022-05-31 at 10 20 00 PM
Screen Shot 2022-06-05 at 6 26 03 PM
Screen Shot 2022-06-05 at 10 20 07 PM

### Notes 
- Unlike the Figma, the front page has a single name field as opposed to separate first and last name fields. This is because the general consensus seems to favor the former over the latter. See [this Slack thread](https://cornelldti.slack.com/archives/C03G3EK5B4G/p1653628926446749) for more details.
- Many of the design changes were accomplished with inline sx styling because I wasn't sure how to factor out the sx styling into a seperate stylesheet, which would probably be a cleaner, more DRY implementation.


frontend/src/modules/Survey/Components/StepBegin.tsx
```diff
@@ -98,48 +46,39 @@ export const StepBegin = ({
...
+            }
+          />
+          <TextField
+            value={email}
+            sx={textInputStyle}
+            variant="standard"
+            placeholder="Cornell NetID"
+            onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+              setEmail(e.target.value)
+            }
```
Coreball: I think we should keep the placeholder as Cornell Email or something, because internally the rest of the code expects emails (especially relevant to sending emails). Unless we want to change to using NetID throughout, which on the dev side at least involves more work.

CollinWoo: I could change it so that "@cornell.edu" is appended to the the end of the target value before setEmail is called. Or is it better just to have it be an email field?

wizhaaa (commented):
Great work on the survey. 

I think the form should have some more form/data validation. Some suggestions I have are 

1) Enforce that the user puts an email in the Net ID field and validate that it is in a valid email format. This will be important later when we want to send emails to the matched students. It may cause an error if the email recipients aren't actually emails. Also, they won't get an email to tell them they were matched. I believe MUI has a prop to pass in to show an error if their input isn't in the form of an email. 

2) Not necessary but would make things easier for the student: catch survey errors beforehand - check that the course they input exists. Maybe call the Cornell course roster API for each course they input and check if it exists as they fill out the form and show the student according to if it exists. Alternatively, could just read the backend response (res.success = false or true) and render the appropriate error screen telling them that their course input is invalid at the end instead of when they fill it out. 

3) Also for courses, I think it may be more accessible if we allow students to put in lower case courses like 'cs 2110' and then we would call .toUpperCase() on their input so it is formatted correctly. Could also strip any white space from the end and beginning of it. 

Not super important, but I feel like some students may use their phone/mobile device to fill out the survey and it is not very mobile friendly as the first page's input fields (name and email) are not visible/tappable.

frontend/src/modules/Survey/Components/StepBegin.tsx
Coreball: Idk

frontend/src/modules/Survey/Components/StepBegin.tsx
```diff
@@ -24,63 +22,13 @@ export const StepBegin = ({
...
-      return
-    }
-    // TODO: change this to some regex magic @Shi Chong
-    if (email === '') {
-      console.log('setting email red')
-      setError(errorEnum.EMAIL)
-      return
-    }
-    setError(errorEnum.NONE)
     gotoNextStep()
```
Coreball: Since the rest of `handleNext` has been removed, you can simplify by just giving the button the `gotoNextStep` function for when it's clicked

frontend/src/modules/Survey/Components/StepBegin.tsx
Coreball: Regarding what @notwz said about validation, if we did want to have them put in their email here the error could just be like "error if it's not empty and doesn't end in @cornell.edu" (technically doesn't catch all invalid emails, but will handle most cases). This way the error doesn't show up before they have typed anything.

Also what if we assume they put in their NetID but someone gives us their email, then we have to remove the @cornell.edu because we don't want netid@cornell.edu@cornell.edu

frontend/src/modules/Survey/Components/StepBegin.tsx
```diff
@@ -98,48 +46,39 @@ export const StepBegin = ({
...
+              setEmail(e.target.value)
+            }
+          />
         </StyledFields>
-        <Box sx={{ marginLeft: 'auto', textAlign: 'center' }}>
-          <IconButton className="next" onClick={handleNext}>
+        <Box
+          sx={{
+            marginLeft: 'auto',
+            textAlign: 'center',
```
Coreball, line 74: Probably a more elegant solution now is to use a column-oriented flexbox to group the button and "Next" instead of using a `<br>` element (bad use of semantic HTML here)

frontend/src/modules/Survey/Components/StepCourse.tsx
```diff
@@ -31,15 +35,19 @@ export const StepCourse = ({
...
+    courses.length === 0 ? 'ABC 1100' : '+ Add another course...'
 
   return (
     <StyledContainer>
       <StyledQuestionText>
         What course(s) would you like to find study partners for?
       </StyledQuestionText>
+      <StyledWarningText>
+        * Please do not submit the cross-listed version of the same course more
+        than once
```
Coreball: Again, I would advocate for less strong wording like "You do not need to submit the cross-listed version of the same course more than once" because it's a strength that we don't need them to do it multiple times, not a flaw on our end that submitting the cross-listed version more than once would break something. I made a comment on the Figma about this just now.

frontend/src/modules/Survey/Components/StepTemplate.tsx
```diff
@@ -45,23 +45,37 @@ export const StepTemplate: FunctionComponent<StepTemplateProps> = ({
...
+              color: 'purple.100',
+            }}
+          >
+            <IconButton
+              className="next"
+              onClick={handleNext}
+              disabled={!isStepValid}
+              sx={{ boxShadow: 3 }}
+            >
+              {stepNumber === totalSteps ? <Check /> : <ArrowForward />}
```
Coreball, line 75: Checkmark and "Finish" are nice!

frontend/src/modules/Survey/Styles/StepCourse.style.tsx
```diff
@@ -10,13 +10,21 @@ export const StyledQuestionText = styled.div`
...
 `
 
+export const StyledWarningText = styled.div`
+  ${h4};
+  font-size: 24px;
+  font-weight: 400;
+  text-align: center;
+  margin: 0px 20%;
+  margin-bottom: 50px;
+`
```
Coreball, line 23: It would be better to move everything over to `<Typography>` in the future, but this can slide for now

Coreball (approved):
Good job, approving this because it does what we need it to do, but there are some comments to look at.

Ideally we would do more to move towards using MUI components directly instead of wrapping them in more components that add complexity and not much value. Replacing the NameField and EmailField with regular TextFields is good progress towards this! However it would also be nice to see the InputFields on the course input step replaced as well, and the InputField file deleted (along with NameField and EmailField) because we aren't using them anymore and don't want to confuse people in the future. It also bothers me a little that the input fields on the course step don't match the styling of the first step.

Then for the course input steps I think you can use onChange and have it update every time the user types something in. This would be a good opportunity to auto-capitalize as they type, so the user doesn't have to worry about "cs 1110" or "CS 1110", whatever they type will show up in the same format that visually matches the placeholder. I believe this would address @notwz's third point in an even more user-friendly way as well.

For calling the course roster API from the survey directly to see if the courses exist and mobile support, I think those can be out of scope for this PR to keep it more streamlined. We should definitely revisit mobile support in the near future though.

Coreball:
Also the overflow looks goofy on Firefox

Screen Shot 2022-06-02 at 22 21 53

frontend/src/modules/Survey/Components/StepBegin.tsx
Coreball: Low priority so I guess it's OK if you don't fix this now

frontend/src/modules/Survey/Components/Survey.tsx
```diff
@@ -6,51 +6,69 @@ import { StepBegin } from 'Survey/Components/StepBegin'
...
 import { Question } from '@core/Types'
+import axios from 'axios'
+import { API_ROOT, STUDENT_API } from '@core/Constants'
 
 export const Survey = () => {
   const [showError, setShowError] = useState(false)
-  const [currStep, setCurrStep] = useState(0)
+  const [currStep, setCurrStep] = useState(1)
+  const [surveySuccess, setSurveySuccess] = useState(false)
+  const [surveyError, setSurveyError] = useState('')
```
Coreball: I think surveySuccess is redundant because if there's an error message that means it was not successful. You could use `useState<string | null>(null)` and check if the error is null or not

frontend/src/modules/Survey/Components/Survey.tsx
```diff
@@ -6,51 +6,69 @@ import { StepBegin } from 'Survey/Components/StepBegin'
...
+    axios.post(`${API_ROOT}${STUDENT_API}/survey`, surveyData).then(
+      (response: any) => {
+        console.log(response)
+        setSurveySuccess(true)
+        setCurrStep(currStep + 1)
+      },
+      (error: any) => {
+        console.log(error)
+        setSurveySuccess(false)
+        setSurveyError(error.message)
```
Coreball: You likely want to use the error message provided in the response, not the message on the error object. For example right now when submitting an invalid course, the error displayed is

> "Error: Request failed with status code 400"

instead of

> "Could not find course data for class CS2114"

which is in the response message. Maybe https://axios-http.com/docs/handling_errors can help

Coreball (commented):
Good work, though there is one user-impacting fix to make regarding the error message. Also remove the old Zing logo from the final step page

Coreball (approved):

PR #59, opened 2022-06-03
Email modal revisions

mluo24:
### Summary 

Because of the issue raised in https://cornelldti.slack.com/archives/C03FG2U1V28/p1653692556464509, as when a React component's state gets modified, the component will indeed refresh, causing a flash, the `ZingModal` component needed to be reworked. The original rational was to not need to remember to add a bunch of extraneous components for styling to make sense, but unfortunately that cannot be accomplished with the way React works. The modal component was instead created to model the [Bootstrap Modal](https://react-bootstrap.github.io/components/modal/) and the [MUI Dialog component](https://mui.com/material-ui/react-dialog/). 

This PR also implements a few stylistic code choices to make code look a bit cleaner. However, not all was implemented (actually, they were implemented independently from Changyuan's concerns), so please take a look at the review.

The following new components were added, in addition to the ZingModal original component:

- [x] ZingModal.Title - underlying component is the `DialogTitle` component, meant to hold the title as well as the close symbol, which can be shown by passing in an onClose component
- [x] ZingModal.Body - underlying component is the `DialogContent` component, can basically hold anything of the Modal body
- [x] ZingModal.Controls - underlying component is the `DialogActions` component, holds the buttons

I left some of the added props that were added in the original PR. Everything else could pretty much be used to style with sx.

### Test Plan 

image
image
image
https://cornelldti.slack.com/archives/C03H6UPNFHC/p1654240862218699

### Notes 

Check the code diff to see the new props of the modal components. New documentation coming soon. Let me know if I missed anything.


Coreball (approved):
Nice, I like how you improved the code in the PR this is branched off from too. One thing which I found odd though is that clicking on the backdrop of the modal no longer dismisses it, was this intended? @hexbabe please approve this one because it is directly relevant to #57

PR #60, opened 2022-06-03
Firebase Cloud logging

wizhaaa:
### Summary 

This PR implements firebase cloud functions logging using firebase's logger SDK.  

See: [https://firebase.google.com/docs/functions/writing-and-viewing-logs](url)

Mostly logs in info and error severities when we hit specific backend endpoints. The purpose of these logs is to be able to debug and catch any errors during live development quickly. 

- [x] cloud function logging for commonly used backend routes 

### Test Plan 

Hit common backend endpoints such as: 
- Submit student survey
- Match students 
- Manually dragging students between groups
- Manually dragging students from unmatched into a group
- Emailing students/matched groups 

Can view logs in the emulator link: [http://localhost:4000](URL)

Sample logs I was able to recreate: 
image
image
image
image
image

### Notes 

Essentially does the same as `console.log` but more specific with severities and more formatted. 

How to use: 
`import { logger } from 'firebase-functions'`

Documentation for logger has `const functions = require('firebase-functions')` and directly replacing that into an import statement does not work as `functions` is undefined. 

`logger.info` or `logger.error` or `logger.warn` etc. depending on your need. 

Also just realized that newline `\n` does not work with this logger. 

### Next 

Not 100% sure where else we need to do logging and/or if this is the best approach to do logging. Just my initial attempt after looking at the documentation and code.

### DIFF
--- a/backend/functions/src/matching/routes.ts
+++ b/backend/functions/src/matching/routes.ts
@@ -20,16 +20,13 @@ router.post('/make', (req, res) => {
   const { courseId, groupSize } = req.body
   ;(groupSize ? makeMatches(courseId, groupSize) : makeMatches(courseId))
     .then((data) => {
-      logger.info('Made matched for course: ' + courseId)
+      logger.info(`Made matched for course: ${courseId}`)
       res.status(200).json({ success: true, data })
     })
     .catch((err) => {
       logger.error(
-        'Failed to match for course: ' +
-          courseId +
-          '. Error message: \n  ' +
-          err +
-          ' \n ---- end of error. \n '
+        `Failed to make matches for course ${courseId}. Error: `,
+        err
       )
       console.log(err)
       res.status(400).json({ success: false, err: err.message })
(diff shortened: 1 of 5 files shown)


backend/functions/src/matching/routes.ts
```diff
@@ -28,8 +39,12 @@ router.post('/make', (req, res) => {
 router.get('/:courseId', (req, res) => {
   const courseId = req.params.courseId
   getGroups(courseId)
-    .then((data) => res.status(200).json(data))
+    .then((data) => {
+      logger.info('Successfully attempted to get course info for ' + courseId)
+      res.status(200).json(data)
+    })
```
Coreball: This hasn't logged during my testing, is this route actually being used?

Coreball (approved):
Looks good! These are the situations in which I would log too, situations where data is being changed. I think you can remove the newlines, it will be ok. I looked into logging as well and discovered that data can be passed in as an object, but I think that's more for log metadata than the actual data.

Just remember that string template literals exist, I think it would make some logging statements in the code easier to read. Prettier won't try to break those into multiple lines either.

backend/functions/src/matching/routes.ts
wizhaaa: I don't believe so either. Just made it just in case we aren't able to load in course data at any point, I guess just being better safe than sorry? 

Honestly, I keep forgetting string template literals exist, thank you. Makes it much cleaner and easier to write. Fixed it literals. 

As for passing objects into logging statements, I didn't really know which objects to pass in, so I didn't pass much in except for errors so far.

PR #61, opened 2022-06-05
Email timestamps backend

wizhaaa:
### Summary 

This PR implements the backend functionality of updating the firebase database with the time when the email was sent for each group. 

Updates the backend endpoint to send email to also update the timestamp field when the email is sent successfully. 

This pull request is the first step toward implementing setting the email sent timestamp. 

- [x] backend route and function
- [ ] front-end hitting the backend route #63 

# *New* Test Plan 

1. Submit student surveys 
2. Goto dashboard and match students. 

3. Hit the endpoint to send emails. Now requires more parameters (must input course Id, group number, and template name)
image

5. Goto localhost:4000 (or emulator site) and check that the `shareMatchEmailTimestamp` field is updated to the correct time. 

image

Alt: 
Pass in other templates results in different timestamp fields/values. 
image

6. Purple circle on hover next to group once timestamp is updated: 
image

### Notes and TODOs
Currently acceptable "template names". These are the same as the template name enum sean implemented, so theoretically shouldn't result in any errors. 
image

## Breaking changes 
send email endpoint requires more parameters. must update for #63

### DIFF
--- a/backend/functions/src/emailing/functions.ts
+++ b/backend/functions/src/emailing/functions.ts
@@ -11,12 +11,12 @@ type timestampsType = { [key: string]: string }
 
 const timestamps: timestampsType = {
   'Share matched results': 'shareMatchEmailTimestamp',
-  'First no match notification': 'checkInEmailTimestamp',
-  'Second no match notification': 'firstNoMatchEmailTimestamp',
-  'Request to add student to group': 'secondNoMatchEmailTimestamp',
-  'Request to add student to group (late)': 'addStudentEmailTimestamp',
-  'Ask to join group': 'lateAddStudentEmailTimestamp',
-  'Check in with groups': 'askJoinGroupEmailTimestamp',
+  'First no match notification': 'firstNoMatchEmailTimestamp',
+  'Second no match notification': 'secondNoMatchEmailTimestamp',
+  'Request to add student to group': 'addStudentEmailTimestamp',
+  'Request to add student to group (late)': 'lateAddStudentEmailTimestamp',
+  'Ask to join group': 'askJoinGroupEmailTimestamp',
+  'Check in with groups': 'checkInEmailTimestamp',
 }
 
 const getTimestampField = (template: string) => {
@@ -29,7 +29,7 @@ const getTimestampField = (template: string) => {
  * @param template is the name of the template of the email being sent. Matched emailTemplates.js names made by sean. Currently is the string value, may change to the variable value in future (less wordy).
  * @result updates database to have email sent timestamp to current time.
  *  */
-export const updateEmailTimestamp = async (
+export const updateEmailTimestamp = (
   courseId: string,
   group: string,
   template: string
@@ -37,7 +37,7 @@ export const updateEmailTimestamp = async (
   const time = admin.firestore.FieldValue.serverTimestamp()
   const timestampField = getTimestampField(template)
   // must be string format -> parse here or when calling function
-  await courseRef
+  return courseRef
     .doc(courseId)
     .collection('groups')
     .doc(group)
@@ -134,7 +134,7 @@ export const sendMails = async (
       data: JSON.stringify(message),
     })
     if (response.status === 202) {
-      await updateEmailTimestamp(courseId, group, template)
+      updateEmailTimestamp(courseId, group, template)
     }
     return response.status
   } catch (error) {
--- a/backend/functions/src/emailing/functions.ts
+++ b/backend/functions/src/emailing/functions.ts
@@ -119,7 +119,7 @@ export const sendMails = async (
   authToken: string,
   courseId: string,
   group: string,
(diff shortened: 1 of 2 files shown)


Coreball:
How do you plan on making this extensible for when there are email timestamps for emails other than the initial matching email, such as check in or request to add student

backend/functions/src/emailing/functions.ts
```diff
@@ -1,4 +1,36 @@
...
+  // must be string format -> parse here or when calling function
+  await Promise.all([
+    groups.forEach((group: string) => {
+      courseRef
+        .doc(courseId)
+        .collection('groups')
+        .doc(group)
+        .update({ shareMatchEmailTimestamp: time })
+    }),
+  ])
```
Coreball: Not sure if the Promises here work, I think it's something more like

```js
await Promise.all(
  groups.map((group: string) =>
    courseRef
      .doc(courseId)
      .collection('groups')
      .doc(group)
      .update({ shareMatchEmailTimestamp: time })
  ),
)
```

Note that I omitted the curly braces and return statement since the group numbers get mapped to Promises directly, which are then given to Promise.all

backend/functions/src/emailing/routes.ts
```diff
@@ -49,4 +48,28 @@ router.post('/send', (req, res) => {
     )
 })
 
+/** @.com/api root/email/timestamp
+ *
+ * @param courseId = string of 6-digit course code
+ * @param groups = string [] of groups to be sent emails
+ *
+ * @yeilds email sent timestamp update in database
```
Coreball, line 69: yeilds

Coreball (commented):
Making it the frontend's responsibility to call the endpoint to send emails and the endpoint to update the timestamps feels a little strange to me, though it does allow us to do one or the other separately if we ever need to. I guess with this setup it would be something like "for every group, send email" and then "update timestamps for all groups" ?

Wondering about how this would work for emails other than the newly matched groups email. Or are we just planning to refactor it later

backend/functions/src/emailing/routes.ts
wizhaaa: Yeab good points. I don't really know tbh. Do we need store timestamps for other types of emails sent? I was thinking that we needed this so the front-end & LSC Staff know which groups they sent emails to or something. If there aren't that many types of emails to sent (matched, check-ins, adding students, asking to join, etc.) then maybe make the timestamp field an enum or something, and then pass that into the function that updates the field according to the enum type? Kind of raw and brute force method though. Or, just have separate functions for each type of shared email if we don't have a lot.

Not 100% sure on the best code design as not 100% sure when/how/what scenarios we would use it and for what purpose. I'll try to think more about it once I finish the email modal functionality.

Coreball: Yeah maybe some sort of mapping between timestamp-type string given to the function and what property in the database it should update could work. I'll think more about it too

backend/functions/src/emailing/functions.ts
```diff
@@ -1,4 +1,50 @@
...
+
+const timestamps: timestampsType = {
+  'Share matched results': 'shareMatchEmailTimestamp',
+  'First no match notification': 'checkInEmailTimestamp',
+  'Second no match notification': 'firstNoMatchEmailTimestamp',
+  'Request to add student to group': 'secondNoMatchEmailTimestamp',
+  'Request to add student to group (late)': 'addStudentEmailTimestamp',
+  'Ask to join group': 'lateAddStudentEmailTimestamp',
+  'Check in with groups': 'askJoinGroupEmailTimestamp',
+}
```
Coreball: I think some of these are not quite right, e.g. `'Second no match notification': 'firstNoMatchEmailTimestamp'`

backend/functions/src/emailing/functions.ts
```diff
@@ -1,4 +1,50 @@
...
+  template: string
+) => {
+  const time = admin.firestore.FieldValue.serverTimestamp()
+  const timestampField = getTimestampField(template)
+  // must be string format -> parse here or when calling function
+  await courseRef
+    .doc(courseId)
+    .collection('groups')
+    .doc(group)
+    .update({ [timestampField]: time })
```
Coreball: Another way of looking at this function could be to think of it as a wrapper around the Firestore call to update the timestamps, in which case you could remove the async and just return the courseRef...update(). Basically taking the parameters, processing them, and returning the Promise that is the database write. Don't think it's necessary though

Coreball (approved):
Looks good! I think the timestamp to field name mapping is slightly off though.

backend/functions/src/emailing/functions.ts
wizhaaa: Oop yeah I think I didn't match them up correctly when I was reformatting it lol. good catch

PR #62, opened 2022-06-05
Various summer updates, Matching algorithm 7 = 3 + 2 + 2

Coreball:
### Summary 

This pull request was supposed to only do a few small things before summer launch but turned into writing more backend code than I expected. It updates things like the title and favicon, but also simplifies get course id logic and more importantly, changes the matching algorithm.

Previously, the matching algorithm formed as many groups of 3 as it could, and the remainder were left unmatched. Now whenever there are at least 2 students, they will always be grouped. This means with 7 students, there will be 1 group of 3 students and 2 groups of 2 students: 7 = 3 + 2 + 2 instead of 7 = 3 + 3 + 1. Groups now index starting at 1!

You should avoid putting so many unrelated changes into one PR like this

- [x] update favicon, title, and description (remove old Zing logo)
- [x] dashboard log out button is a text button
- [x] simplify backend code for mapping course catalog name to course id
- [x] change matching algorithm to always make groups when at least 2 students
- [x] group numbers start at 1 (not just a visual change)
- [x] hide full emails when ending in `@cornell.edu` (show NetID)
- [x] hide unfinished Match button on dashboard cards
- [x] snackbar alert thing primarily for network errors, usable everywhere

### Test Plan 

Favicon and title update:

Screen Shot 2022-06-05 at 16 12 28

Dashboard log out button:

Screen Shot 2022-06-05 at 16 26 38

New matching algorithm:

https://user-images.githubusercontent.com/22627336/172074893-f000d09b-60aa-4eec-ba82-6fbca17210e6.mov

Use this snackbar when there's no other way of telling the user there was an error:

Screen Shot 2022-06-06 at 23 37 35

Tested that the simplified backend code for mapping course catalog name to course id still works by submitting student surveys. POST requests with invalid course names are now rejected immediately.

### Breaking Changes 

- group numbers are now indexed starting at 1 in the database
- firebase-functions package upgraded from 3.6.1 to 3.21.2


wizhaaa (approved):
Tested on my end and everything seems to be working. 

 I like stripping the @cornell.edu for displaying students IDs as well.

PR #63, opened 2022-06-06
Email modal functionality 

wizhaaa:
# Summary 

This PR connects the backend functionality of the email modals to the frontend. This allows the LSC staff to select groups  and send emails to them. 

Includes a failure flow to allow staff to re-login (refresh auth token) one time in case their auth token expired. 

This pull request is one of the final step towards implementing email modal functionality

- [x] can send emails
- [x] correctly display if email send err
- [x] update timestamp (?) (later/concurrent PR)

# Test Plan 

## *Updated* Video (includes timestamp) 
Link:
https://streamable.com/ewoife

*Old* Video Link: 
https://streamable.com/a5c8v8

## Success Flow:
1. After Pressing "Send x Emails" button, a snackbar appears with link to the LSC shared inbox. 
image

2. Separate emails for each group appearing in their inbox (as they are included in recipient list), but does not show up in the "Sent" folder. 
image

3. Email is "sent" by the LSC email, students are able to reply directly. 
image

4. Shows purple icon/timestamp 
Screen Shot 2022-06-09 at 12 45 47 AM

## Failure Flow: 
1. User is not permitted to send emails on behalf of LSC, or their login/auth token has expired for the session. 
Prompts them to login in (refresh the auth token and instantly attempts to log in one more time. 
image

2. If the try again succeeded, it would do the success flow. 
Else displays:
image

### Notes 

I have added Changyuan as an owner of the LSC email so perhaps he can also test this out in case there exists any code I put that makes it work only for me.


frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -1,25 +1,92 @@
...
+   * @returns the student emails for a given group in a string array.
+   */
+  const groupEmails = async (group: any) => {
+    const rcpts: string[] = ['lscstudypartners@cornell.edu']
+    const members = group.memberData
+    await members.forEach((mem: any) => {
+      rcpts.push(mem.email)
+    })
+    return rcpts
+  }
```
Coreball: Don't think the async/await is necessary here, can also be as simple as
```js
return ['lscstudypartners@cornell.edu', ...group.memberData.map(mem => mem.email)]
```

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -1,25 +1,92 @@
...
+      const emailSubject = 'Study Partners!'
+      const emailItems = { emailSubject, emailRcpts, emailBody }
+      const emailingResult = await sendEmail(emailItems)
+      if (emailingResult === false) {
+        failure = true
+        console.log(`"made it here" : ${failure}`)
+      }
+    }
+    return failure
+  }
```
Coreball: From what I read as I was looking up for-await trying to figure out what this new syntax is, I think it's meant for asynchronous iterators? I think this would be the cleanest to read if `Promise.all` was used instead so the emails can all be sent at the same time. Also then you can change sendEmail promise to reject if the emailing fails, and put a catch on the `Promise.all`, which will reject if any of the constituent promises reject.

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -101,6 +237,28 @@ export const EmailModal = ({
     }
   }
 
+  const SelectTemplates = () => {
+    return (
+      <>
+        {' '}
```
Coreball: lmao the space character `{' '}`, is your editor configured to delete trailing whitespace

frontend/src/modules/Emailing/Components/Emailing.tsx
```diff
@@ -135,24 +135,24 @@ export const Emailing = () => {
...
-    SUCC -> 201 
-    FAIL -> 401  */
-export const sendEmail = async (
-  emailItems: any,
-  setEmailSent: (arg: boolean) => void
-) => {
+/** Returns (bool)
+    SUCC -> true 
+    FAIL -> false  */
+export const sendEmail = async (emailItems: any) => {
```
Coreball, line 139: Might be more idiomatic to throw errors, especially because this is an async function so you can do things like
```js
sendEmail(emailItems)
  .then(/* something */)
  .catch(error => /* handle email failure */)
```
Or maybe not because failure to send is expected to happen sometimes

Would like other people's input here @hexbabe @CollinWoo @mluo24 (look at rest of PR for more context)

Coreball (commented):
According to the PR desc you are still planning to clean this up but I left some preliminary feedback for some parts. Looks good so far, especially having the detailed modal pages that show up when there are errors! I can't test the success mode myself because I do not have access to the LSC inbox, but the Streamable video showcased it very well. Otherwise it works as I'd expect when I don't have the permissions. Also prevents developers from accidentally emailing a bunch of people in prod

Lmk when to actually approve

frontend/src/modules/Core/Constants/Theme.ts
```diff
@@ -209,6 +209,9 @@ theme = createTheme(theme, {
             borderColor: theme.palette.essentials[50],
           },
         },
+        textPrimary: {
+          color: theme.palette.purple[50],
+        },
```
Coreball: To add element-specific styling to emailSentAction in the email success snackbar, put this as a prop for the Button (don't need to style the Snackbar itself):
```js
sx={{ color: 'purple.50' }}
```

frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -41,6 +43,28 @@ export const EditZing = () => {
...
+        onClick={() =>
+          window.open(
+            'https://outlook.office.com/mail/lscstudypartners@cornell.edu/',
+            '_blank'
+          )
+        }
+      >
+        View in Inbox
+      </Button>
+    </React.Fragment>
```
Coreball: Can remove the <React.Fragment> because only one element is being returned rn

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -56,6 +134,76 @@ export const EmailModal = ({
...
+            *This will automatically attempt to send the email one more time.{' '}
+          </em>
+        </Typography>
+        <Button
+          onClick={() => {
+            adminSignIn().then(() => handleEmailSend())
+          }}
+          sx={{ position: 'absolute', bottom: '20%' }}
+        >
+          Try Again{' '}
```
Coreball: Are the random space characters to be inserted in the HTML in this component and the others needed

Coreball (approved):
Looks really good! It's super satisfying to watch the emailing feature come together. It's taken so much developer effort to make it working and this PR finally makes it work end to end. Great job!

frontend/src/modules/EditZing/Components/EmailModal.tsx
wizhaaa: Yeah, I don't think they are. Just my editor settings... which I should probably turn off.

frontend/src/modules/Core/Constants/Theme.ts
wizhaaa: Now I feel dumb trying to change the snackbar props lol

PR #64, opened 2022-06-10
Catch Class Roster API 404 Errors

Coreball:
### Summary 

This pull request provides better error messages for the student survey when invalid courses are submitted. Previously, we only returned a custom error message when the class could not be found in a successfully returned array of classes. Now a custom error message is returned when Class Roster can't even find the subject or course levels.

- [x] better survey error messages by catching 404 errors

### Test Plan 

Tested classes such as ABC 1100, CS 8000, CS 1111, CS 3110, and CS 31110 for SU22
- ABC 1100 - subject does not exist
- CS 8000 - course level does not exist
- CS 1111 and CS 3110 - classes are not offered in SU22 (or don't exist)
- CS 31110 and cs 3110 - improper class format

### Notes

Only catching 404 errors because they are expected behavior. Other errors will be passed through and logged to the Cloud Functions console.

### DIFF
--- a/backend/functions/src/course/get_course_id.ts
+++ b/backend/functions/src/course/get_course_id.ts
@@ -22,7 +22,15 @@ async function getCourseId(
   }
 
   const url = constructUrl(courseCatalogName, roster)
-  const response = await axios.get(url)
+  const response = await axios.get(url).catch((error) => {
+    if (error.response.status === 404) {
+      throw new Error(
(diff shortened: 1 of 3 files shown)


wizhaaa (approved):
Nice! Correctly displays the error messages at the end if you put class levels that don't exist and if the course subject doesn't exist either.

PR #65, opened 2022-06-11
Displaying other email timestamps 

wizhaaa:
# Summary 

This PR extends off #43 in that we will also display email sent and tooltips for 2 more templates: Check In and Request to add students. This means that all templates that are able to be selected and sent by staff are now supported. 

### Note for future: 
Not sure if we will/need to display for other/all email types? Currently has 7 templates which would mean 7 dots, which may or may not look funny or be necessary? Well, might be necessary to keep track of all the different emails they send. Not sure. 

This pull request is the first step towards implementing displaying email sent info/timestamps. 

- [x] implemented check in email tooltip
- [x] implemented add student tooltip
- [ ] remaining email templates not implemented 

# Test Plan 
General Flow: 
1. Survey
2. Match 
3. Send an email for each template 
4. Refresh (want to automatically re-render but idk best way on how to since the current state changes we do don't cause re-renders) 
5. Purple dots should appear on next to the group name & number. 
6. Check In tooltip:
Screen Shot 2022-06-11 at 12 04 04 PM
7. Req add student tooltip: 
image
8. Extra: In database correctly sets timestamps as null initially: 
image

### DIFF
--- a/backend/functions/src/matching/functions.ts
+++ b/backend/functions/src/matching/functions.ts
@@ -121,7 +121,7 @@ async function makeMatches(courseId: string) {
       updateTime: new Date(),
       shareMatchEmailTimestamp: null,
       checkInEmailTimestamp: null,
-      addStudentEmailTimestamp: null, // timestamp for option of "share matching results"
+      addStudentEmailTimestamp: null,
     })
   }
   for (let i = 0; i < studentDataDoubles.length; i += 2) {
@@ -134,7 +134,7 @@ async function makeMatches(courseId: string) {
       updateTime: new Date(),
       shareMatchEmailTimestamp: null,
       checkInEmailTimestamp: null,
-      addStudentEmailTimestamp: null, // timestamp for option of "share matching results"
+      addStudentEmailTimestamp: null,
     })
   }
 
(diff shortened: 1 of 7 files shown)


backend/functions/src/matching/functions.ts
```diff
@@ -119,7 +119,9 @@ async function makeMatches(courseId: string) {
       memberData: newGroup,
       createTime: new Date(),
       updateTime: new Date(),
-      shareMatchEmailTimestamp: null, // timestamp for option of "share matching results"
+      shareMatchEmailTimestamp: null,
+      checkInEmailTimestamp: null,
+      addStudentEmailTimestamp: null, // timestamp for option of "share matching results"
```
Coreball: I think the comment is misplaced with the wrong line of code

Coreball (approved):
Looks good. It would be nice to maybe rework the timestamp format so the different tooltips can be added using a loop of some kind instead of a new component for each one. This would also involve changing the email update part too. Not sure if it's a high priority to fix, this works perfectly fine too. However changing it down the line would be harder once there's real user data in the system. Good job

Coreball:
Also for having the purple dots show up right after emailing, you can update the local state of the group with a `new Date()` for the timestamp as the email request completes. This approximates the timestamp and when the user visits the page in the future the timestamp will be the real one recorded in the database.

Kind of like how when dragging students between groups, they are updated locally in addition to updated in the database. When the page gets refreshed or visited again the groups are read from the db.

PR #66, opened 2022-06-13
Course ID format: SU22-358546

Coreball:
### Summary 

This pull request changes the course ID format to be of the form `SU22-358546`, which includes the roster the course is associated with and the course number. Using only the course number would cause the same class in two different semesters to be blended together in our current database, since the Class Roster API maps CS 2110 in SU22 and FA22 to be 358546.

- [x] changed course IDs to be of the form `SU22-358546`
- [x] minor type fixes
- [x] display roster on edit page

### Test Plan 

The URLs look like this now. Also added the semester to the title.

Screen Shot 2022-06-12 at 16 52 57

### Notes 

```json
"SU22-358546": {
  "roster": "SU22",
  "courseNumber": "358546",
  "names": ["CS 2110", "ENGRD 2110"],
  "..."
}
```

### Breaking Changes 

- course ID are different format!

### DIFF
--- a/backend/functions/src/course/functions.ts
+++ b/backend/functions/src/course/functions.ts
@@ -17,7 +17,7 @@ async function getDataForStudents(emails: string[]) {
   return Promise.all(emails.map((email) => getStudent(email)))
 }
 
-async function getCourseInfo(courseId: any) {
+async function getCourseInfo(courseId: string) {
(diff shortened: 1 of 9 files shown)


wizhaaa (approved):
Awesome work, I tested most of the features and they work for me and are updated correctly in the database as well.

PR #67, opened 2022-06-14
Implement Landing Page

CollinWoo:
### Summary 
This pull request implements the landing page design shown [here](https://www.figma.com/file/9zueLCvKT7BBMYi8Dt8Ukp/Zing-LSC?node-id=3259%3A227513). See http://localhost:3000.

- [x] add button to redirect students to the survey
- [x] implement design to match the Figma as close as possible

### Test Plan 

Changes were mostly visual so testing was done by comparing the Figma to the actual implementation. Screenshots comparing the two are below:

#### Figma
Screen Shot 2022-06-13 at 11 10 25 PM

#### Implementation
Screen Shot 2022-06-13 at 11 06 15 PM


frontend/src/modules/Core/Constants/Theme.ts
```diff
@@ -195,6 +195,15 @@ theme = createTheme(theme, {
...
         },
+        outlinedPrimary: {
+          borderColor: theme.palette.common.white,
+          backgroundColor: theme.palette.purple.main,
+          color: theme.palette.common.white,
+          '&:hover': {
+            borderColor: theme.palette.common.white,
+            backgroundColor: theme.palette.purple[120],
+          },
+        },
```
Coreball, line 206: Now this is very cool, great for reusability

frontend/src/modules/Home/Components/Home.tsx
```diff
@@ -1,79 +1,56 @@
...
   return (
     <StyledBackground>
       <StyledContainer>
         <StyledLeftPanel>
           <StyledWhiteActionText>
-            Connect students, create groups
+            <Typography variant="h2" fontWeight={'600'}>
+              Connect students. <br />
+              Create groups.
+            </Typography>
```
Coreball: I think you should consolidate instances where an element only has one child element, into just one element, especially when the components aren't used elsewhere. This collapsing of elements removes redundant HTML and puts all the styling on that element in one place.

For example,
```js
<Box sx={{ a: x }}>
  <Box sx={{ b: y }}>
    ...
  </Box>
</Box>
```
is cleaner as
```js
<Box sx={{ a: x, b: y }}>
  ...
</Box>
```

Coreball (commented):
The landing page looks really good Colin, it's very faithful to the design. I especially liked how you added the outlined primary button style to the Theme file because it may be used in other contexts, and how you made the "I'm a Student" button use `component={Link} to="/survey"` instead of using the imperative `history.push` API. It would also be good to use the `SURVEY_PATH` constant found in Routes.ts.

I think that this would also have been a great opportunity to completely rewrite the Home component, since not much is shared in common with the old home page, visually. Sometimes working on existing code is harder and messier than starting over from scratch, where everything can be written fresh with new best practices, not needing to worry about updating existing code.

In this case you'll still want to remove the unused styles from Home.style.tsx and consolidate some CSS as mentioned above. Some like `StyledHeaderText` are unused now and some like `StyledContainer` don't have a real purpose.

I would also suggest adding `component="h1"` to give the heading text on the page the semantic HTML component of `<h1>`, which is better for accessibility than the second-level heading of `<h2>` w/o a primary heading. Sometimes the screen design is not conducive to a proper h1, but I think it works here.

Coreball (approved):
Looks good, nice work

PR #68, opened 2022-06-17
Survey accessibility v1 

wizhaaa:
### Summary

This PR is working towards adding accessibility to the survey. Utilize the WAVE extension to audit our survey screens. Main features added are: 

- [x] aria labels 
- [x] fieldsets 
- [x] input labels
- [x] solidify color 
- [x] make the first step work on mobile/smaller screens 
- [x] h1 heading structure (maybe not necessary) 
- [x] main tag (maybe not necessary) 

This pull request is the first step toward implementing survey accessibility. 

### Test Plan

1. Currently the home page on an iPhone 12 Pro:
image

2. Changed to (removed the left panel completely on smaller screen widths):
image

3. Audit on the different steps using WAVE:
image
image
image
image

### Notes 

No real design to work off, just googled form design best practices and implemented how I saw fit. 

ALSO: 
For the courses step, once you put in a course, the text has a failing color contrast. I could make it darker pink? 

AND, I have put the label on the last, most current text field and not the previous ones the user already put in. So that breaks the Input Label audit, but would look messy with labels on each one.

### DIFF
--- a/frontend/src/modules/Survey/Components/StepBegin.tsx
+++ b/frontend/src/modules/Survey/Components/StepBegin.tsx
@@ -1,7 +1,14 @@
 import React from 'react'
-import { IconButton, Box } from '@mui/material'
+import {
+  IconButton,
+  Box,
+  TextField,
+  InputLabel,
+  FormControl,
+  Input,
+  FormHelperText,
+} from '@mui/material'
 import { ArrowForward } from '@mui/icons-material'
-import { TextField } from '@mui/material'
 import {
   StyledContainer,
   StyledLeftPanel,
@@ -12,6 +19,7 @@ import {
   StyledTitleWrapper,
   StyledHeaderText,
   StyledWelcomeText,
+  StyledErrorText,
 } from 'Survey/Styles/StepBegin.style'
 import { StepBeginProps } from 'Survey/Types'
 
@@ -44,29 +52,42 @@ export const StepBegin = ({
           <StyledHeaderText>Hi,</StyledHeaderText>
           <StyledWelcomeText>Find study partners!</StyledWelcomeText>
         </StyledTitleWrapper>
+
         <StyledFields>
+          <InputLabel htmlFor="user name"> Name: </InputLabel>
           <TextField
-            placeholder="Name"
+            id="user name"
             variant="standard"
+            aria-label="user name"
+            placeholder="Martha E. Pollack"
+            // variant="standard"
             sx={textInputStyle}
             value={name}
             onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
               setName(e.target.value)
             }
           />
+          <InputLabel htmlFor="user email"> Email: </InputLabel>
           <TextField
+            id="user email"
+            variant="standard"
+            aria-label="user email"
             value={email}
             sx={textInputStyle}
-            variant="standard"
             type="email"
             onBlur={() => setIsValidEmail(validEmail.test(email))}
-            placeholder="Cornell Email"
+            placeholder="mep22@cornell.edu"
             onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
               setEmail(e.target.value)
             }
             error={!isValidEmail}
-            helperText={isValidEmail ? ' ' : 'Invalid Email'}
           />
+          <FormHelperText id="email-helper-text" sx={{ fontColor: 'red' }}>
+            {' '}
+            <StyledErrorText>
+              {isValidEmail ? ' ' : 'Invalid Email'}
+            </StyledErrorText>
+          </FormHelperText>
         </StyledFields>
         <Box
           sx={{
@@ -81,6 +102,7 @@ export const StepBegin = ({
             onClick={gotoNextStep}
             disabled={name === '' || email === '' || !isValidEmail}
             sx={{ boxShadow: 3 }}
+            aria-label="Next button"
           >
             <ArrowForward />
           </IconButton>
--- a/frontend/src/modules/Survey/Components/StepBegin.tsx
+++ b/frontend/src/modules/Survey/Components/StepBegin.tsx
@@ -7,6 +7,7 @@ import {
   FormControl,
   Input,
   FormHelperText,
+  Typography,
 } from '@mui/material'
 import { ArrowForward } from '@mui/icons-material'
 import {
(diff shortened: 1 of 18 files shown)


frontend/src/modules/Survey/Components/StepBegin.tsx
```diff
@@ -40,53 +47,75 @@ export const StepBegin = ({
...
+          <StyledTitleWrapper>
+            <StyledHeaderText>Hi,</StyledHeaderText>
+            <StyledWelcomeText>Find study partners!</StyledWelcomeText>
+          </StyledTitleWrapper>
+
+          <StyledFields>
+            <InputLabel htmlFor="user name">
+              {' '}
+              <StyledLabelText> Name: </StyledLabelText>{' '}
+            </InputLabel>
```
Coreball: The random extra spaces {' '} are odd

frontend/src/modules/Survey/Components/StepBegin.tsx
```diff
@@ -40,53 +47,75 @@ export const StepBegin = ({
...
+                setEmail(e.target.value)
+              }
+              error={!isValidEmail}
+            />
+            <FormHelperText id="email-helper-text" sx={{ fontColor: 'red' }}>
+              {' '}
+              <StyledErrorText>
+                {isValidEmail ? ' ' : 'Invalid Email'}
+              </StyledErrorText>
+            </FormHelperText>
```
Coreball: What is wrong with the existing helper text that is part of MUI? May be able to style it if needed. Also `fontColor: red` is not constrasty enough

frontend/src/modules/Survey/Components/StepBegin.tsx
```diff
@@ -40,53 +47,75 @@ export const StepBegin = ({
...
+              className="next"
+              onClick={gotoNextStep}
+              disabled={name === '' || email === '' || !isValidEmail}
+              sx={{ boxShadow: 3 }}
+              aria-label="Next button"
+            >
+              <ArrowForward />
+            </IconButton>
+            <br />
+            Next
```
Coreball: I would suggest using `aria-labelledby` instead of `aria-label` here because MDN web docs says "If there is visible text that labels an element, use [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) instead." Would need to put the "Next" in its own element in this case, which is probably better anyway because styling with the `<br />` is not good practice

Was mainly thinking about this for StepTemplate but here too

frontend/src/modules/Survey/Components/StepRadio.tsx
```diff
@@ -19,31 +20,42 @@ export const StepRadio = ({
...
+        <StyledQuestionText>{question.question}</StyledQuestionText>
+        <StyledRadioButtonsWrapper>
+          {showError ? (
+            <StyledErrorWrapper>
+              <StyledErrorIcon />
+              <StyledErrorText>
+                Please select an item to continue
+              </StyledErrorText>
+            </StyledErrorWrapper>
+          ) : null}
```
Coreball: I guess this and the associated code can actually be removed now since the next button is simply disabled if they don't have anything selected yet

frontend/src/modules/Survey/Components/StepRadio.tsx
```diff
@@ -19,31 +20,42 @@ export const StepRadio = ({
...
+          <StyledContainer>
+            <StyledFieldSet>
+              <legend>
+                {/**
+                 * TODO
+                 * do we want a legend here or no?  */}
```
Coreball: Nah

frontend/src/modules/Survey/Components/StepCourse.tsx
```diff
@@ -48,38 +51,48 @@ export const StepCourse = ({
...
+              value={course}
+              onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+                updateCourse(index, e.target.value)
+              }
+              error={
+                !validCourseRe.test(course)
+                  ? 'Must be of the form ABC 1100'
+                  : ''
+              }
+            />
```
Coreball: It would be ideal if these could be replaced with regular MUI TextFields too instead of the InputFields, which are unnecessary wrappers around TextField. Styling them to not be pink would be easier then because all the MUI styling is available. They were already removed on StepBegin but not here; I'm not sure but some business logic may have to be tweaked depending on onChange and onBlur

Coreball (approved):
Thanks so much for working on survey accessibility Will. I'm reviewing this PR late at night so you have the most time to work on it. Great job making the survey more accessible. This task was something I had stronger opinions on for some parts and you have done good job following accessibility guidelines. The adapting of the first page to a smaller screen size is nice to see too. The darker purple on the survey looks really nice. I think if the course fields are replaced with regular text fields they won't be as pink anymore. I think labeling the last text field only is okay on the courses step, because it's the primary one the user will interact with. The others are basically just showing the data they already entered. Approving this because everything works but look at my suggestions

frontend/src/modules/Survey/Components/StepBegin.tsx
wizhaaa: https://mui.com/material-ui/react-text-field/#accessibility 

This is what I referenced and if I use the helperText prop of the TextFeild component, I end up with a broken aria label error, so I did it this way.

frontend/src/modules/Survey/Styles/Survey.style.tsx
```diff
@@ -1,16 +1,30 @@
...
   height: 100%;
-  background-image: url(${bg1});
-  background-size: cover;
-
   display: flex;
   justify-content: center;
   align-items: center;
+
+  // blob overlay
+  &:before {
```
Coreball, line 17: Missing other curly brace, as seen through syntax extension in VSCode

Screen Shot 2022-06-17 at 18 04 09

frontend/src/modules/Survey/Components/StepBegin.tsx
Coreball: Hm, I think FormHelperText is already part of the TextField and what you might be looking at in the documentation is for "If you are composing the component" which we are not

Coreball:
Surprised some changes were done manually instead of merging in the other branch

frontend/src/modules/Survey/Components/StepBegin.tsx
Coreball: I think this is because the id of the TextField was `"user email"`, with the space. Removing the space fixed it and I was able to get the broken aria label working.

PR #71, opened 2022-06-22
validate emailz

wizhaaa:
### Summary 

This PR validates emails on the backend as well. 

Currently, I did validate email => then(add student survey response). Might have a cleaner implementation? 

Just have to make sure if it doesn't pass the email validation, it also doesn't just add the student data to the db. 

- [x] cornell email regex on email sent in
- [x] returns error if error 

### Test Plan 

1. postman send surveys 
image

2. Get error msg
image

3. check db if added or not 
4. check with valid email form too

### DIFF
--- a/backend/functions/src/student/functions.ts
+++ b/backend/functions/src/student/functions.ts
@@ -24,7 +24,7 @@ async function removeStudentFromCourse(
   }
 }
 
-const validateEmail = async (email: string) => {
+const validateEmail = (email: string) => {
   const emailRegex = /^\w+@cornell.edu$/
   // if true => valid email. if false => invalid email.
   const validEmail = emailRegex.test(email)
@@ -90,6 +90,16 @@ const addStudentSurveyResponse = async (
     courseId: course.courseId,
     groupNumber: -1,
   }))
+
+  const emailRegex = /^\w+@cornell.edu$/
+  // if true => valid email. if false => invalid email.
+  const validEmail = emailRegex.test(email)
+  console.log(validEmail)
+  if (!validEmail) {
+    const e = new Error('Invalid Email Form')
+    e.name = 'processing_err'
+    throw e
+  }
 
   // First, update the [student] collection to include the data for the new student
   const studentUpdate = studentRef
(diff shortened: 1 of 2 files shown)


backend/functions/src/student/functions.ts
```diff
@@ -24,6 +24,17 @@ async function removeStudentFromCourse(
...
 }
 
+const validateEmail = async (email: string) => {
+  const emailRegex = /^\w+@cornell.edu$/
+  // if true => valid email. if false => invalid email.
+  const validEmail = emailRegex.test(email)
+  console.log(validEmail)
+  if (validEmail === false) {
+    throw new Error('Invalid Email Form')
+  }
```
Coreball: I think we should avoid using `console.log` in backend code now because we can use the new logger functions. Is this important to log, without any context?

I feel like `if (validEmail === false)` can be written more idomatically

Coreball (commented):
Thanks for validating the emailz Will. I do think there exists a cleaner implementation, but I like your test plan.

I think it would be preferable to have the validate email check in `addStudentSurveyResponse` because that's where invalid courses are detected too, just for consistency. For invalid courses line 44 in functions.ts throws an error and it's caught in the `.catch` in routes.ts. This may change in the near future though.

`validateEmail` probably shouldn't be asynchronous since it doesn't have a good reason to. There are not any long running or async calls within it, and the nested promises are not nice to look at.

Also, I don't think HTTP 500 is the right status code to send when an email is invalid. That means there was an internal server error, in this case HTTP 400 Bad Request makes more sense. That's also what is returned when a course does not exist.

backend/functions/src/student/functions.ts
wizhaaa: Ok yeahhh, I don't think my brain was working tbh. Thanks for the tips, I think I implemented it a bit better now?

Put it inside the singular function, the condition, and removed the console logging (I forgot to remove it initially) Lmk if there's still anything else.

Coreball (approved):
LGTM

PR #72, opened 2022-06-23
Partial survey success

Coreball:
### Summary 

This pull request adds partial survey success. Previously, if one course was not found in the class roster for the specified semester we would abort processing the survey submission entirely. Now it will continue processing with the classes that it could find, and return a response with data on what classes were added successfully and which weren't added. This data is used on the survey confirmation screen on the frontend to let the user know what was added and what wasn't, so they have a chance to fix their errors if needed.

- [x] survey can now partially succeed

### Test Plan 

Sample request and response. Notice how the response also combines ENGRD 2110/CS 2110 so it's extra clear to the user that cross-listed classes were recognized and processed correctly, if they submitted the same course under different titles.

Screen Shot 2022-06-22 at 22 22 52

Happy path

Screen Shot 2022-06-22 at 23 12 09

Complicated case with multiple added and failed, cross listed classes

Screen Shot 2022-06-22 at 23 11 32

Courses added: none

Screen Shot 2022-06-22 at 23 13 05

Continue to show real errors (used https://httpstat.us/ to test, simulates if the Class Roster API went down)

Screen Shot 2022-06-22 at 23 38 11

### Notes 

TypeScript target changed to `es2020` so I can use [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled)


backend/functions/tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "outDir": "lib",
     "sourceMap": true,
     "strict": true,
-    "target": "es2019",
+    "target": "es2020",
```
hexbabe, line 10: ?

hexbabe (approved):
Thanks for the hot fix. Check one comment for clarification. LGTM

backend/functions/tsconfig.json
Coreball: > TypeScript target changed to es2020 so I can use [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled)

from PR description, does that make sense

PR #73, opened 2022-06-24
Update email template for newly matched

Coreball:
### Summary 

Update email template for newly matched students as requested. Also minor CSS for overflow styling.

- [x] update newly matched email template

### Test Plan 

Sent it to myself and got the new email

### Notes

Consider automation, but it would require LSC knowledge of HTML unless we turn Zing LSC into a full fledged text editor


wizhaaa (approved):
looks gud

PR #74, opened 2022-06-25
Survey confirmation redirect and updated college list

CollinWoo:
### Summary 

This pull request implements the updated list of colleges described [here](https://lscstudypartnersdti.slack.com/archives/C02A94AS65U/p1656018680602909) and the survey redirect link shown [here](https://www.figma.com/file/9zueLCvKT7BBMYi8Dt8Ukp/Zing-LSC?node-id=3383%3A227669). 

- [x] Update the list of colleges
- [x] Change text above the image on the survey begin step to match Jen's request
- [x] Add a redirect link to lsc.cornell.edu from the survey confirmation page

### Test Plan 

Changes were mostly visual so testing was done by comparing the Figma and written descriptions to the actual frontend. Screenshots of the frontend implementation are below:

Screen Shot 2022-06-24 at 8 36 24 PM
Screen Shot 2022-06-24 at 8 55 34 PM
Screen Shot 2022-06-24 at 8 55 38 PM
Screen Shot 2022-06-24 at 8 36 02 PM
Screen Shot 2022-06-24 at 8 38 01 PM

### Notes
* In addition to the requested changes, there are minor styling adjustments to the confirmation screen and the radio buttons on the survey


Coreball (commented):
Good job simplifying CSS, I love to see it. The scrolling colleges now does feel a little strange but I guess it's harder to avoid now that there are more colleges in the list and they take up more space.

One thing I would like to suggest (may want to bring this up with designers too) is what the buttons should be like on the confirmation screen when the warning is shown. Personally, I feel like it'd make sense to have a second button going back to the first page of the survey since the text says to "resubmit the form if needed". If we just have a button linking back to the LSC website in this case then they might press it and have to find the form link again. One button makes sense if the survey was submitted and processed successfully, since then they'll only need to go back to the LSC website.

I also feel like the text should not be "back to home" because the LSC website is a different site than the website this is on. Even "back to LSC home" is clearer.

Coreball (approved):
Looks good! Though I do wonder if the radio buttons could just be wider instead of being 2 or 3 lines with scrolling

Coreball (approved):
Good

Coreball (approved):

PR #76, opened 2022-06-26
Responsive Survey 

wizhaaa:
# Summary 

I just wanted to make the survey a bit more responsive and reduce the amount of clipping/display funkiness on smaller screens, particularly mobile screens less than 900 pixels wide. 

The following applies to widths < 900px: 
- [x] full width and height for each step
- [x] no more clipping (from my testing) 
- [x] wider text fields for the beginning step 
- [x] scroll down to see prev/next buttons (no y scroll for radio buttons on smaller width) 
- [x] success screen content wraps rather than clips 

## Test Plan 

On dev tools and mobile view, everything works on the smallest screen size of the preset custom responsive dimensions. 

### Video: 
ignore audio lmao

https://user-images.githubusercontent.com/46132945/175795683-6c1f7913-ed46-4206-b7f5-3c6207d5d217.mov

### Screenshots: 
1. Home 
image
2. Course Selection:
image
3. College Selection
image
image
4. Year
image
5. Final Screen
image
image

### DIFF
--- a/frontend/src/modules/Survey/Styles/StepBegin.style.tsx
+++ b/frontend/src/modules/Survey/Styles/StepBegin.style.tsx
@@ -1,6 +1,6 @@
-import styled, { css } from 'styled-components'
+import styled from 'styled-components'
 
-import { colors, h1, h2, h3, StyledComponent } from '@core'
+import { StyledComponent } from '@core'
 
 import teamPic from '@assets/img/teamwork.svg'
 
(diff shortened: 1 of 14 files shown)


Coreball (commented):
This is good and it works well on my end! But since we're moving away from Styled Components I feel like we should avoid adding more to it, because it makes it harder to continue the migration to fully using MUI's system later.

For example, instead of using CSS media queries you could use MUI's breakpoints, which you can learn more about at https://mui.com/system/basics/#responsive-values and https://mui.com/material-ui/customization/breakpoints/ . It provides an easy way to change how things look based on screen size and 900px already matches the standard `md` breakpoint. I would also suggest replacing text with `Typography` when possible.

This also looks a bit weird near the bottom:

Screen Shot 2022-06-25 at 19 50 32

frontend/src/modules/Survey/Styles/StepBegin.style.tsx
```diff
@@ -63,17 +61,25 @@ export const StyledFields = styled.div`
...
+  }
 `
 
 export const StyledWelcomeText = styled.text`
   ${h2};
   font-weight: 300;
   color: ${colors.black};
+  @media (max-width: 900px) {
+    font-size: 2rem;
+  }
```
Coreball: Can some of this file be deleted now?

Coreball (approved):
Nice work! Tested on a real phone with the preview URL and it's a lot nicer than the pre-responsive version.

frontend/src/modules/Survey/Styles/StepBegin.style.tsx
wizhaaa: 🚮 yah I deleted what I think was safe so far to delete just now

PR #77, opened 2022-07-01
Update email template for check in

Coreball:
### Summary 

Update email template for check in as requested. Thinking about changing how email templates work, but nothing yet because other things are higher priority. In the meantime, manual process.

- [x] update check in email template

### Test Plan 

Previewed email in send email modal


wizhaaa (approved):
LgMt

PR #78, opened 2022-07-01
Emailing individual students frontend implementation

wizhaaa:
# Summary 

PR to allow emailing individual students. 

Currently, utilizes the same email modal as we use to send group emails. As of now, it is working for me. 

Possible changes: 
- conditionally render different template/modal content based on selectedGroups/selectedStudents etc... 
- or just make a new modal (more work but simpler)

### Test Plan 

1. Selecting students
image
2. Selecting both students + groups is disabled
image
3. Student Emails populate the `To:` field 
image
4. Has correct number of emails to be sent: 
image
5. Checkbox handling overflows:
image
6. New style:
image
image

### DIFF
--- a/frontend/src/modules/EditZing/Components/EmailModal.tsx
+++ b/frontend/src/modules/EditZing/Components/EmailModal.tsx
@@ -57,27 +57,22 @@ export const EmailModal = ({
    * @throws error if there are any errors in sending the email
    */
   const sendIndividualEmails = async () => {
-    try {
-      await Promise.all(
-        selectedStudents.map((student) => {
-          const emailRcpts = [student]
-          const emailBody = getBody(selectedTemplate, courseNames.join(', '))
-          const emailSubject = 'Study Partners!'
-          const emailItems = {
-            emailSubject,
-            emailRcpts,
-            emailBody,
-            courseId,
-            groupNum: -1,
-            selectedTemplate,
-          }
-          return sendEmail(emailItems)
-        })
-      )
-    } catch (err) {
-      console.log(`Failed to send individual emails`)
-      throw new Error('Failed to send individual emails')
-    }
+    await Promise.all(
+      selectedStudents.map((student) => {
+        const emailRcpts = [student]
+        const emailBody = getBody(selectedTemplate, courseNames.join(', '))
+        const emailSubject = 'Study Partners!'
+        const emailItems = {
+          emailSubject,
+          emailRcpts,
+          emailBody,
+          courseId,
+          groupNum: -1,
+          selectedTemplate,
+        }
+        return sendEmail(emailItems)
+      })
+    )
   }
 
   /**
@@ -86,28 +81,23 @@ export const EmailModal = ({
    * @throws error if there are any errors in sending the email
    */
   const sendGroupEmails = async () => {
-    try {
-      await Promise.all(
-        selectedGroups.map((group) => {
-          const emailRcpts = groupEmails(group)
-          const emailBody = getBody(selectedTemplate, courseNames.join(', '))
-          const emailSubject = 'Study Partners!'
-          const groupNum = group.groupNumber.toString()
-          const emailItems = {
-            emailSubject,
-            emailRcpts,
-            emailBody,
-            courseId,
-            groupNum,
-            selectedTemplate,
-          }
-          return sendEmail(emailItems)
-        })
-      )
-    } catch (err) {
-      console.log(`Failed to send group emails`)
-      throw new Error('Failed to send group emails')
-    }
+    await Promise.all(
+      selectedGroups.map((group) => {
+        const emailRcpts = groupEmails(group)
+        const emailBody = getBody(selectedTemplate, courseNames.join(', '))
+        const emailSubject = 'Study Partners!'
+        const groupNum = group.groupNumber.toString()
+        const emailItems = {
+          emailSubject,
+          emailRcpts,
+          emailBody,
+          courseId,
+          groupNum,
+          selectedTemplate,
+        }
+        return sendEmail(emailItems)
+      })
+    )
 
     // emailStudents().then((res) => {
     //   if (res === false) failure = true
--- a/frontend/src/modules/EditZing/Components/EmailModal.tsx
+++ b/frontend/src/modules/EditZing/Components/EmailModal.tsx
@@ -51,13 +51,6 @@ export const EmailModal = ({
     ]
   }
 
-  const individualEmails = (students: string[]) => {
-    return [
-      'lscstudypartners@cornell.edu',
-      ...students.map((email: string) => email),
-    ]
-  }
-
   /**
    * promise that sends emails to each individual student.
    *
@@ -65,8 +58,8 @@ export const EmailModal = ({
    */
   const sendIndividualEmails = async () => {
     await Promise.all(
-      selectedStudents.map(() => {
-        const emailRcpts = individualEmails(selectedStudents)
+      selectedStudents.map((student: string) => {
+        const emailRcpts = [student, 'lscstudypartners@cornell.edu']
         const emailBody = getBody(selectedTemplate, courseNames.join(', '))
         const emailSubject = 'Study Partners!'
         const emailItems = {
--- a/frontend/src/modules/EditZing/Components/EmailModal.tsx
+++ b/frontend/src/modules/EditZing/Components/EmailModal.tsx
@@ -15,6 +15,7 @@ import { adminSignIn } from '@fire/firebase'
 
 export const EmailModal = ({
   selectedGroups,
+  selectedStudents,
   isEmailing,
   setIsEmailing,
   courseNames,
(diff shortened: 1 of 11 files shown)


frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -58,7 +59,7 @@ export const EmailModal = ({
   const sendGroupEmails = async () => {
     let failure = false
 
-    await Promise.all(
+    await Promise.all([
```
Coreball: Putting the contents of the Promise.all in an array without flattening changes how it works. Instead of `Promise.all([Promise, Promise, Promise])` it is now `Promise.all([[Promise, Promise, Promise], []])`. The difference is subtle, but in the second case Promise.all only sees arrays of something and doesn't wait for the promises in those arrays to complete.

The quick fix is to just `.flat()` everything before handing it to Promise.all, but I would encourage other ways of doing this such as updating the function name to be more than sendGroupEmails because it sends emails to individual students now too, and changing it so the groupNum maybe just isn't even sent. Right now groupNum: -1 seems inaccurate if individual students being emailed are actually in a group.

Coreball: It would also be better to use Errors more because they work well with Promises, since throwing an Error causes the Promise to [reject](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject) which indicates failure. Also has the benefit of being able to throw the error through multiple levels of functions back to where it can be caught and handled appropriately, instead of having to do the `failure = true` to pass the result from sending the emails up through this function.

frontend/src/modules/EditZing/Components/StudentGrid.tsx
```diff
@@ -64,12 +65,32 @@ export const StudentGrid = ({
...
+          className={classes.paper1}
+        >
+          <Checkbox
+            color="secondary"
+            checked={selected}
+            onChange={handleChecked}
+            sx={{
+              display: selected || isHovering ? 'flex' : 'none',
+            }}
+          />
```
Coreball: Checkbox positioning is funky, take a look at https://www.figma.com/file/9zueLCvKT7BBMYi8Dt8Ukp/Zing-LSC?node-id=1131%3A45021

Coreball (commented):
Good work so far Will, but I think this PR needs a bit more time in the oven. The way the checkbox ripple area is a stretched oval doesn't fit well with the rest of the app and how it pops in is a little distracting too. I suppose styling could be handled in another PR, I'm just concerned it would take us a long time to finish styling it, when we could just knock it out in this PR too since the component is already being added.

Re: "the design is just copying the same flow for group emailing", I don't think it is because for emailing groups you can either use the drop down to auto-select groups or select them manually, then the button changes to Email Selected which opens the modal. For emailing individual students right now the email modal opens through the auto-select dropdown instead of the Email Selected button? The Email Selected button is not even shown when students are selected. I think this makes it confusing for the user because the entry point for emailing students and groups is different. Things in the dropdown should do similar things, which in my mind is auto-selecting students/groups that match certain criteria such as not having the newly matched email sent yet.

Any thoughts on my Notion writeup I made? I wasn't sure earlier, but now I think that showing the Email Selected button when either groups or students are selected is good, then it can be disabled state when students and groups are selected at the same time. Right now there are no safeguards against emailing students+groups at the same time. I'm not sure if emailing students+groups at the same time is a feature we're adding later, but right now I don't think any emails work like that?

The student emails should also be listed on the modal because right now it just says To:

wizhaaa:
@Coreball Okay, so I think I fixed most of the concerns. I believe I threw errors inside the promises and caught them and handled them better this time. Although, there are a lot of layers to the emailing code/functions. I may look back on it in the future and think about how to make it more concise. As perhaps throw errors after the graph API call instead of other places. But not 100% sure yet. 

I fixed the checkbox to mirror the Figma a bit more accurately. I think though we might want to make Group Cards/student cards with fixed width + heights. Right now, their widths are flexible, grow and shrink, which may cause the checkbox icon button to overlap with the student's name and also results in clipping issues. Like perhaps use flex display instead of MUI Grid? And/or limit the overall width of the entire page to be like, say 1440px. But, this is more of an admin utility which they may want to take up as much space on a page and display as much info as possible but with Grid, it can only display 4 (I think?) group cards per row, but they will be super stretched out as opposed to however many per row their display can handle. Also, just my perspective that naming StudentGrid and GroupGrid is kind of weird because they aren't a grid per se but a grid item. 

Also, I am leaving the `selectedStudents` as a string array and making group number = -1 as I am too lazy right now to do the code to add individual student timestamps. I think it would probably be something like an array of 
`{ email: string, groupNum: num/string }` instead of just the email. But, I guess which emails would they send individual students? I feel like just the no-match email? 

Let me know if there is anything else I can improve on.

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -72,30 +95,28 @@ export const EmailModal = ({
...
-          if (res === false) failure = true
-        })
+        return sendEmail(emailItems)
       })
     )
 
-    return failure
+    // emailStudents().then((res) => {
+    //   if (res === false) failure = true
+    // })
```
Coreball: 🚮

frontend/src/modules/EditZing/Components/StudentGrid.tsx
```diff
@@ -1,56 +1,19 @@
...
-
-  & .${classes.paper2} {
-    display: flex;
-    flex-direction: column;
-    gap: 2px;
-    margin-top: 2px;
-  }
-`
-)
+import { Checkbox, Box, Typography } from '@mui/material'
```
Coreball: Nice work moving to more MUI-native

frontend/src/modules/EditZing/Components/StudentGrid.tsx
```diff
@@ -80,17 +65,43 @@ export const StudentGrid = ({
...
+                </Typography>
+                <Typography sx={{ fontWeight: '400', fontSize: '0.875rem' }}>
+                  {student.email.replace('@cornell.edu', '')}
+                </Typography>
+                <Typography sx={{ fontWeight: '400', fontSize: '0.875rem' }}>
+                  {student.year}
+                </Typography>
+              </Box>
+
+              <Checkbox
```
Coreball: You can try adding `disableRipple` prop to this checkbox because the current ripple looks odd b/c of how close it is to the edge of the card in the designs

Coreball (commented):
Looks good Will! The flow for sending individual students emails is more like sending group emails now and the checkboxes for students look a lot nicer. I think the error handling is done correctly for the most part, yeah it could be made more concise in the future.

Agree that the group cards/student cards might want to be fixed height/width, because the growing/shrinking of them looks weird especially when the window is resized. I'm not a huge fan of MUI Grid, I think it was made before CSS Grid became a thing. Regular Flexbox would probably work too. Agree that naming them StudentGrid and GroupGrid does not make much sense any more, because StudentGrid doesn't have a grid inside it at all.

Yeah it's ok to leave groupNumber = -1 for right now. Timestamp-related things will be a new task.

Had a weird experience when testing where it showed me the final error page for sending emails but I still got an email in my inbox. Also, is the lsc email going to be included as a recipient on emails sent to individual students? Or only to groups?

Screen Shot 2022-07-02 at 16 08 36

Screen Shot 2022-07-02 at 16 37 15

Tooltip positioning for when student submitted survey is slightly messed up too, fix this and will approve

Screen Shot 2022-07-02 at 16 13 58

wizhaaa:
image

- [x] Added LSC as recipient of email as well. 
- [x] Fix checkbox 
- [x] fixed sizes for group/student cards

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -51,15 +51,22 @@ export const EmailModal = ({
...
    * promise that sends emails to each individual student.
    *
    * @throws error if there are any errors in sending the email
    */
   const sendIndividualEmails = async () => {
     await Promise.all(
-      selectedStudents.map((student) => {
-        const emailRcpts = [student]
+      selectedStudents.map(() => {
+        const emailRcpts = individualEmails(selectedStudents)
```
Coreball, line 69: Does this send the right number of emails?

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -50,14 +51,43 @@ export const EmailModal = ({
...
+
+  /**
+   * promise that sends emails to each individual student.
+   *
+   * @throws error if there are any errors in sending the email
+   */
+  const sendIndividualEmails = async () => {
+    await Promise.all(
+      selectedStudents.map(() => {
+        const emailRcpts = individualEmails(selectedStudents)
```
Coreball: Does this send the right number of emails? Probably don't want to include all of the students' emails in the recipients list because they will see each other when they are meant to be emailed individually

frontend/src/modules/EditZing/Components/GroupCard.tsx
```diff
@@ -122,17 +125,27 @@ export const GroupGrid = ({
             }}
           />
         </Box>
-        <Grid container spacing={2}>
+        <Box
+          sx={{
+            display: 'grid',
+            gridTemplateColumns: 'repeat(auto-fit, minmax(112px, max-content))',
+            gap: '16px',
```
Coreball, line 132: I think `gap: 2` may also work here if the default gap size is 8px (uses theme spacing)

Coreball (approved):
Gonna approve this for now but please confirm/disconfirm what may be multiple emails sent? That's the impression I got from the code but I have not tested it. Fixed size cards are nice, in the future maybe one way we could make it look better is centering everything on the screen, or providing some min/max size. Also the shadow transition for the student cards feels long compared to the group card

frontend/src/modules/EditZing/Components/EmailModal.tsx
wizhaaa: Ah yes, my fault on this.  Good catch

wizhaaa:
Okay fixed the emailing to actually be individual (my brain wasn't working when i wrote that lol) 

and then made the group grids with CSS grid so that its centered and that unmatched is the same width as the total width of all group cards combined for a row. 

I don't know why it seems like the student cards hover animations appear slower? I don't have any animations/transisitons, maybe just placebo/size of shadow causes it to appear slower.

PR #79, opened 2022-07-02
Notes backend

Coreball:
### Summary 

This pull request is the first step towards implementing student notes on a per-class basis. It sets up the backend and database to handle tracking a notes string for each class a student is in. Since the new route uses the student router that also has to handle public survey requests (until we add student sign in) I had to add a new middleware that secures all routes except for `/survey`.

Also included is factoring out `getStudent` and adding a `Student` type (finally!) to the backend. Not everywhere has been migrated yet. Also prefer `Timestamp.now()` instead of `serverTimestamp()` in some situations going forward because it's better to be consistent for things updated in the same function execution instead of when exactly things are written to the backend, since those times may differ by fractions of a second. Also updated `utils/add-test-data.ts` to use `@cornell.edu` emails.

- [x] implemented backend route for updating notes
- [x] existing backend routes that get student data get note modification timestamp in right format
- [x] other code improvement

POST `/student/notes` request to update notes for a student for a particular class:
```json
{
  "email": "a@cornell.edu",
  "courseId": "SU22-358526",
  "notes": "notes test"
}
```

New group membership type for students. `notesModifyTime` is sent as a stringified Date object, need to convert it back to a Date on the frontend like `submissionTime` and others. `notes` will always exist as a field, empty notes are represented by empty strings:
```json
"groups": [
  {
    "notesModifyTime": "2022-07-02T07:24:17.155Z",
    "notes": "notes test",
    "courseId": "SU22-358526",
    "groupNumber": 1
  },
  {
    "notesModifyTime": "2022-07-02T07:22:10.923Z",
    "notes": "",
    "courseId": "SU22-358546",
    "groupNumber": -1
  }
],
```

### Test Plan 

Success input/output for POST `/student/notes`

Screen Shot 2022-07-02 at 00 39 33

Error outputs for POST `/student/notes`

Screen Shot 2022-07-02 at 00 44 48

Screen Shot 2022-07-02 at 00 44 37

Output from GET `/course/students/SU22-358526`

Screen Shot 2022-07-02 at 00 25 01

Output from POST `/matching/make` (one unmatched student, but same if there are matched students)

Screen Shot 2022-07-02 at 00 25 43

### Notes 

Spent a lot of time figuring out why my changes weren't being built. Old files remained in the `lib` folder and I didn't even know what was going on because there were no error messages, the code I was writing just didn't do anything. Turns out that trying to import service account as an ES6 import and typing it as ServiceAccount changes the folder structure of the output, nestling everything in src/lib... I didn't notice because it was using my old compiled files!

### Breaking Changes

- Fields for `notes` and `notesModifyTime` are required to exist or loading the edit page will fail. Need to update database accordingly

### DIFF
--- a/backend/functions/src/config.ts
+++ b/backend/functions/src/config.ts
@@ -1,6 +1,4 @@
 import admin from 'firebase-admin'
-import { config } from 'dotenv'
-config()
 
 const serviceAccount: string = require('../service-account.json')
 
--- a/backend/functions/src/course/functions.ts
+++ b/backend/functions/src/course/functions.ts
@@ -1,21 +1,6 @@
-require('dotenv').config({ path: '../.env' })
-
 import { db } from '../config'
+import { getStudentsData } from '../student/functions'
 const courseRef = db.collection('courses')
-const studentRef = db.collection('students')
-
-async function getStudent(email: string) {
-  const snapshot = await studentRef.doc(email).get()
-  if (!snapshot.exists) throw new Error(`Student ${email} does not exist`)
-  const result: any = snapshot.data()
-  result.email = email
-  result.submissionTime = result.submissionTime.toDate()
-  return result
-}
-
-async function getDataForStudents(emails: string[]) {
-  return Promise.all(emails.map((email) => getStudent(email)))
-}
 
 async function getCourseInfo(courseId: string) {
   const snapshot = await courseRef.doc(courseId).get()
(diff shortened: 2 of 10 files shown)


wizhaaa (approved):
Tested on my end + working. I was thinking about what would happen if we stored the notes in the course > groups > members. That would change the data structure to be not just a string but an object that also contains the email + notes. So keeping it in the student collection is probz good and need less change to db + code

Coreball:
Yeah I have thought of doing it that way too. I think keeping it in student collection is beneficial because it's easier to gather all notes for a student if that's ever needed. Makes it harder to gather all notes associated with a course without querying the students, but if we need all notes for a course we will almost certainly need to query student data too

PR #80, opened 2022-07-06
Notes frontend

wizhaaa:
# Summary

This PR implements the front-end side of the student notes feature. With this PR, the notes feature should be finished. 

- [x] note modals for each student
- [x] alerts for note saves + save failures

### Current Implementation of Edit + Checkbox icon. 
image

### Notes
An appropriate snack bar alert appears when the note is successfully saved and if a save error occurs. 

## Test Plan 

Optional: `npm run populate` + send own survey in. 
Output: Multiple students in multiple groups for multiple classes have their correct notes saved and updated. 

2. Create a note in another class (PHYS 2213) 
image

3. Create a note for another class (MATH 1920) but in an unmatched group
image

4. Check DB + reload to see if note is saved for each class 
image

And yes, notes are still saved. 

5. Check if updated notes are saved 
image
6. Saved in DB:
image


frontend/src/modules/EditZing/Components/NotesModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+import LoadingButton from '@mui/lab/LoadingButton'
+import SaveIcon from '@mui/icons-material/Save'
+import CloseIcon from '@mui/icons-material/Close'
+
+import { API_ROOT } from '@core/Constants'
+import axios from 'axios'
+
+const NotesModal = (props: any) => {
+  const { courseId } = useParams<{ courseId: string }>()
+  const { open, student, handleClose, setSaved, setNotSaved } = props
```
Coreball: Props should be typed. Also suggest deconstructing the props object directly in the function parameters like other components

frontend/src/modules/EditZing/Components/NotesModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const [note, setNote] = useState(student.notes)
+  const [saving, setSaving] = useState(false)
+  const email = student.email
+
+  // getting notes for student
+  useEffect(() => {
+    const savedNote = student.groups.find((g: any) => g.courseId === courseId)
```
Coreball: `student.groups` is of type `GroupMembership[]`. This would be solved automatically by typing the props of this component. Also add `notes` and `notesModifyTime` to that type definition

frontend/src/modules/EditZing/Components/NotesModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+    setSaving(false)
+    handleClose()
+  }
+
+  return (
+    <Modal
+      open={open}
+      onClose={handleClose}
+      sx={{ display: 'flex', justifyContent: 'center', alignItems: 'center' }}
+    >
```
Coreball: @mluo24 made a ZingModal component that's in use in other parts of the app like the email modal. How does it compare to using a plain MUI Modal here?

frontend/src/modules/EditZing/Components/StudentCard.tsx
```diff
@@ -41,6 +49,25 @@ const StudentCard = ({
...
+        autoHideDuration={3000}
+        onClose={() => setSaved(false)}
+        message="Notes saved."
+      />
+      <Snackbar
+        open={notSaved}
+        autoHideDuration={3000}
+        onClose={() => setNotSaved(false)}
+        message="Notes failed to save."
+      />
```
Coreball: Might want to make this visually distinct if it isn't already

frontend/src/modules/EditZing/Components/StudentCard.tsx
```diff
@@ -79,10 +106,12 @@ const StudentCard = ({
...
                 title={
-                  'Requested study partner ' +
-                  (submissionTime.getMonth() + 1) +
-                  '/' +
-                  submissionTime.getDate()
+                  <Typography sx={{ fontWeight: '500', fontSize: '13px' }}>
+                    Requested:{' '}
+                    <Typography sx={{ fontWeight: '900', fontSize: '18px' }}>
+                      {submissionTime.getMonth() + 1}/{submissionTime.getDate()}
+                    </Typography>
```
Coreball: Not gonna lie, the first time I saw this I was confused because I thought it said 7/6 as in seven sixths or like score 7 out of 6 instead of a date.

I think consistency with how Miranda did the tooltips for the group dot timestamps and the designs for the tooltips would be better (current tooltips for groups and students use default MUI styling. Perhaps add a style override in the theme to make all the tooltips look the same?)

frontend/src/modules/EditZing/Components/StudentCard.tsx
```diff
@@ -115,8 +144,34 @@ const StudentCard = ({
                 position: 'absolute',
                 right: '1px',
                 top: '1px',
+                '&:hover': {
+                  transform: 'scale(1.15)',
+                },
```
Coreball: Cool but kind of weird IMO. Wonder what other people think?

frontend/src/modules/EditZing/Components/StudentCard.tsx
```diff
@@ -115,8 +144,34 @@ const StudentCard = ({
...
+                style={{
+                  width: '24px',
+                  height: '24px',
+                  position: 'absolute',
+                  right: '-1px',
+                  top: '22px',
+                  cursor: 'pointer',
+                  display: isHovering ? '' : 'none',
+                }}
+                alt=""
```
Coreball: Why does IconButton not work here

Coreball (commented):
Good work so far, here are my thoughts (some are design-related):
1. We probably need to do a better job of communicating because I think Annie and I talked about some designs relating to this but maybe didn't document it anywhere except for notes on my Figma page
2. I think the notes icon should always show up on the student card. It should be visually distinct (e.g. filled pencil icon) when notes are nonempty for the student, to let the administrator easily know which students have special notes attached to them without needing to hover over every one. (Color variation can reinforce the nonempty notes icon, but it can't be the only differentiation)
3. Related to the above, the positioning of the notes icon under the checkbox would be weird if the notes icon was always there but the checkbox only showed up on hover. The most natural place I can think of is trailing behind the student name, but this also makes it wider.
4. Also related to the above, this would shrink the student name so it's not full width of the student card. This is fine because later student email timestamp dots will need to be there too. Don't think need to worry about the tooltip interfering with side icons if everything's done correctly because it should only show up on hover.
5. Currently not a fan of replacing the checkbox with a mail icon because there's not a common pattern of what a mail icon's selected state is (vs a checkbox: checked = selected)
6. I like the loading button usage for when notes are saving, very cool and prevents wondering why the button no click
7. Feedback for notes saving is nice too!

frontend/src/modules/EditZing/Components/StudentCard.tsx
wizhaaa: Yeah, I was also wondering how could I use Icon Button here? As the notes icons in the Figma file aren't actually mui icons so I just imported it as an image.

Coreball: I haven't tested it but does the Icon Button not let you use the react-component-ed version of the image as its children? In the examples it's like `<IconButton><SomeIcon /></IconButton>` and I thought it wouldn't be limited to only putting their icons as children of the component

wizhaaa: Yeah, it is possible. I just changed it to IconButton. Do we want the ripple effect though? I think the checkbox doesn't use it. I just used Box (with component prop set to button) as it was easier to style. Also removed transformation on hover.

frontend/src/modules/EditZing/Components/NotesModal.tsx
wizhaaa: Yeah, I mean I think a regular modal is just simpler to use.

frontend/src/modules/EditZing/Components/NotesModal.tsx
wizhaaa: Yep fixed the typing too

frontend/src/modules/EditZing/Components/NotesModal.tsx
wizhaaa: Doned

frontend/src/modules/EditZing/Components/StudentCard.tsx
wizhaaa: Yeah I mainly changed it because sometimes the tooltip would get on top of the edit notes icon. But I've changed it back to what it was.

frontend/src/index.css
```diff
@@ -23,6 +23,18 @@ code {
...
+input[type='reset'] {
+  background: none;
+  color: inherit;
+  border: none;
+  padding: 0;
+  font: inherit;
+  cursor: pointer;
+  outline: inherit;
+}
+
```
Coreball: Is this still needed or can the global styling be removed now?

frontend/src/modules/EditZing/Components/NotesModal.tsx
mluo24: how, you're writing more lines of code ;-;

frontend/src/modules/EditZing/Components/StudentCard.tsx
```diff
@@ -57,13 +101,12 @@ const StudentCard = ({
             fontFamily: 'Montserrat',
             fontWeight: '700',
             fontSize: '14',
-            boxShadow: isHovering
-              ? '4px 4px 8px rgba(0, 0, 0, 0.3)'
-              : '0px 2px 5px rgba(205, 156, 242, 0.2)',
+            boxShadow:
+              isHovering && !selected
+                ? '4px 4px 8px rgba(0, 0, 0, 0.3)'
+                : '0px 2px 5px rgba(205, 156, 242, 0.2)',
```
Coreball, line 131: Adding `transition: 'box-shadow 0.1s'` is enough to make the transition speed match that of groups

frontend/src/modules/EditZing/Types/Student.ts
```diff
@@ -9,6 +9,8 @@ export interface Student {
 export interface GroupMembership {
   courseId: string
   groupNumber: number
+  notes: string
+  notesModifyTime: string
```
Coreball: The `notesModifyTime` should be stored as a Date object since that's how other timestamps are represented on the frontend right now. Update the function `responseStudentToStudent` at the bottom of this file to do the conversion.

frontend/src/modules/EditZing/Components/NotesModal.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const { courseId } = useParams<{ courseId: string }>()
+  const [saving, setSaving] = useState(false)
+  const [localVal, setLocalVal] = useState('')
+  const email = student.email
+
+  // getting notes for student
+  useEffect(() => {
+    const savedNote = student.groups.find((g) => g.courseId === courseId)?.notes
+    setLocalVal(savedNote || '')
+  }, [studentNotes, setStudentNotes, courseId, student])
```
Coreball: What is this supposed to do?

frontend/src/modules/EditZing/Components/StudentCard.tsx
```diff
@@ -33,6 +36,21 @@ const StudentCard = ({
...
+  const handleCloseNotes = () => setOpenNotes(false)
+
+  const [studentNotes, setStudentNotes] = useState('')
+
+  useEffect(() => {
+    const savedNote = student.groups.find(
+      (g: GroupMembership) => g.courseId === courseId
+    )?.notes
+    setStudentNotes(savedNote || '')
+  }, [student, courseId])
```
Coreball: I actually don't think keeping student notes as local state for the StudentCard is a good idea because it will be out of sync with the students the top level component sees. If we were to use notes elsewhere on this page, the changes here would not be reflected. I would suggest lifting the note-updating logic up to the top and passing down a function to update the notes for a particular student.

Coreball (commented):
Good work but I think it doesn't quite work exactly as how I would expect, see this video recording for an example:

https://user-images.githubusercontent.com/22627336/179342980-24f2ed6b-2c50-4b42-bc98-6558aeff9e73.mov

The whole local state thing is kind of confusing, especially with how useEffect is being utilized.

frontend/src/modules/EditZing/Components/StudentCard.tsx
wizhaaa: Okay, I reverted it back to a single state. However, this method has the effect of showing the filled notes icon when closing but not explicitly saving. It also saves the value of the canceled note until refreshed. One potential solution I could try is to pass in two different closing methods, one that simply sets the note state back to what it originally was if the user doesn't save, or just handle close regularly if the user saves. But, that doesn't work that well either actually because the stored notes don't get updated on the dashboard level. So may just leave it as is?

Coreball (commented):
I would approach the notes state problem the following way. Remove student notes as "local state" from StudentCard (maybe not actually because of how the component structure works, read the rest of this review for details). Tracking local state here to display the filled notes icon or not could be out of sync with the values of the student notes fields that are present in the top-level student data held in EditZing. Create a function in EditZing that updates notes in the top-level data structure on the frontend. Pass that function down all the way to the NotesModal.

Have local state in the NotesModal (needed to represent the text field, might actually be in the parent component StudentCard but it's intended for the NotesModal). That local state is initially set to the student's relevant notes. When the text field is edited that local state changes. If the user saves the notes, POST the backend and call the function that was passed down from EditZing to update the top-level data structure with the new notes. Updates to that top-level data structure will decide whether the notes icon is filled or not. If the user cancels, set the local state back to the student's original notes, which is whatever is in the top-level data structure. That way when the modal is opened again, it's like they never edited the local state.

This is the general idea. It might require some tweaking, like moving the value that represents the NotesModal state back into the StudentCard because of where the handler for canceling the modal is defined. Let me know if you have any questions.

frontend/src/modules/EditZing/Components/StudentCard.tsx
Coreball: See new review comment

frontend/src/modules/EditZing/Components/StudentCard.tsx
```diff
@@ -121,6 +183,52 @@ const StudentCard = ({
...
+                display: !studentNotes && isHovering ? '' : 'none',
+              }}
+              onClick={handleOpenNotes}
+              color="secondary"
+            >
+              <img
+                src={notesIcon}
+                style={{ width: '24px', height: '24px' }}
+                alt="edit student note"
+              />
```
mluo24: Instead of using images, look into using an [SvgIcon](https://mui.com/material-ui/icons/#svgicon) component.

frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -285,8 +286,6 @@ export const EditZing = () => {
       .catch((error) => displayNetworkError(error.message))
   }
```
mluo24, line 287: so uh, what's going on here?

frontend/src/modules/Core/Constants/Theme.ts
```diff
@@ -270,6 +270,11 @@ theme = createTheme(theme, {
         },
       },
     },
+    MuiTooltip: {
+      styleOverrides: {
+        tooltip: {},
+      },
+    },
```
mluo24: What's the purpose of this?

frontend/src/modules/EditZing/Components/StudentCard.tsx
```diff
@@ -33,6 +39,38 @@ const StudentCard = ({
...
+        courseId: courseId,
+        notes: modalNotes,
+      })
+      .then(() => {
+        updateNotes(student.email, modalNotes) // Share the changes to rest of the page
+        setShowNotesSaveSuccess(true)
+      })
+      .catch(() => setShowNotesSaveFailure(true))
+    setIsNotesSaving(false)
+    setOpenNotesModal(false)
```
mluo24: Is this await not redundant? If you want to do it the await way, I think you can wrap the whole thing in a try catch and move all the elements around accordingly afterwards (i.e. catch into the catch block, then in the try.

frontend/src/modules/EditZing/Components/StudentCard.tsx
```diff
@@ -45,6 +83,32 @@ const StudentCard = ({
...
+        message="Notes saved."
+      />
+      <Snackbar
+        open={showNotesSaveFailure}
+        autoHideDuration={3000}
+        onClose={() => setShowNotesSaveFailure(false)}
+        message="Notes failed to save."
+        ContentProps={{
+          style: {
+            background: '#ff3868',
```
mluo24: use `bgcolor: 'error.main'`. tbh this is optional, as the color isn't exactly the same, but that's part of the theme and it'll look cleaner than just a random hex val

mluo24 (approved):
Nice work! I like your new iterations of this feature, and I really appreciate the little details that went into feedback for user interaction elements (saving shows a loading button, if it's successful or not it will show in a popup). I'm approving this because it works, and I don't see anything wrong with the implementation really, but do make note of my comments.

frontend/src/modules/EditZing/Components/StudentCard.tsx
wizhaaa: Yeah idrk how to use svgs. Should I first convert the image into a svg and then put the path and stuff as the children of the SvgIcon?

frontend/src/modules/EditZing/Components/EditZing.tsx
wizhaaa: I believe Changyuan added a catch for any general errors

frontend/src/modules/Core/Constants/Theme.ts
wizhaaa: I'll delete that. Was trying to change the max width of it so it didnt overlap with the edit icons

frontend/src/modules/EditZing/Components/StudentCard.tsx
wizhaaa: I believe Changyuan did this function ? Unless I did it and forgot

frontend/src/modules/EditZing/Components/StudentCard.tsx
mluo24: yep, that's right. if you downloaded it from Figma, I think you can download it directly as an SVG. You can either copy and paste the path directly or import it as a component.

frontend/src/modules/EditZing/Components/EditZing.tsx
mluo24: whoops, I meant to comment on the comment that said `// this is not efficient, you can just update the groups locally` @Coreball

frontend/src/modules/EditZing/Components/StudentCard.tsx
mluo24: @Coreball ? lol

frontend/src/modules/EditZing/Components/EditZing.tsx
Coreball: See https://github.com/cornell-dti/zing-lsc/pull/65#issuecomment-1153049016. It's better to update the timestamp locally to approximate what was changed on the backend instead of making a second network request getting all of the data again. Like when dragging and dropping students between groups, the groups are updated locally and the request is made to save. It doesn't make the request and then reload all the data again. But @notwz just merged in that PR right after making the commits lol https://github.com/cornell-dti/zing-lsc/blob/e7a9ee1f665d6acd31eb9e01c77397c5b94e81af/frontend/src/modules/EditZing/Components/EditZing.tsx#L128-L150

frontend/src/modules/EditZing/Components/StudentCard.tsx
Coreball: The only change I made to this section of code was adding `updateNotes(student.email, modalNotes)`. Otherwise it is essentially unchanged from the previous iteration, save for some variable/function renaming: https://github.com/cornell-dti/zing-lsc/blob/e126870bb3f0340690862ec284d37d998bc9c4bc/frontend/src/modules/EditZing/Components/NotesModal.tsx#L25-L40

frontend/src/modules/EditZing/Components/StudentCard.tsx
wizhaaa: How is this possible? bgcolor isn't a prop or property right? And setting content props style background to error.main doesnt work either.

frontend/src/modules/EditZing/Components/StudentCard.tsx
wizhaaa: ye i mean i'll just remove the async from it all together

frontend/src/modules/EditZing/Components/StudentCard.tsx
mluo24: Oh wait, I didn't look at the code carefully enough. Instead of style, you need to use sx.

frontend/src/modules/EditZing/Components/StudentCard.tsx
Coreball: Removing the async/await here broke the loading button on the modal because now it doesn't wait for the API call to finish before changing `isNotesSaving` back to false. The modal disappears instantly after pressing the save button.

PR #82, opened 2022-07-07
survey first page styling

wizhaaa:
### Summary 

This PR changes so the first step of the survey is centered vertically on the right panel. 

Also changes the left panel text to match the fulling spelling out LSC. 

Also removes the min width of the radio buttons which doesn't fit on smaller screen sizes. 

### Test Plan 

My regular width (1600px): (changed to strategies)
image
On 75% zoom: 
image

Adds space:
image


Coreball (approved):
Looks good. Hope I did not miss anything else this time

PR #83, opened 2022-07-09
Rename student email script

Coreball:
### Summary 

This pull request adds a command line script for renaming student emails. It copies the student info into the new document in the students collection and updates all references in courses (both when they are unmatched and when they are in a group) to use the new email identifier. Also updated `add-test-data.ts` script to use the new list of colleges.

- [x] added rename student email script
- [x] update add-test-data script to use new colleges

### Test Plan 

Tested on the emulator and used it on the Firestore database.

### Notes 

Enough code here to make an endpoint for it but not doing that yet, because not needed at the moment


wizhaaa (approved):
Nice

PR #84, opened 2022-07-10
Track survey timestamp per class

Coreball:
### Summary 

This pull request modifies the survey timestamp to be tracked per class of a universal submission time for all classes the student is in.

This is beneficial because it allows for greater detail how many times the student submitted the survey, e.g. did they submit both classes at once or did they submit the survey for a class and come back a week later to submit for the other class. It is especially important for usage across semesters because if a return student submitted a new survey, the submission timestamp for the previous semester would be wiped out. This PR prevents that by tracking the survey timestamp on a per-class basis.

Also included are improvements to type definitions for both the frontend (`ResponseStudent` and similar) and the backend (`FirestoreStudent` and similar). Types that start with `Response` are intended to represent what data the backend sends to the frontend. Dates are sent as strings and converted into Date objects on the frontend. Types that start with `Firestore` represent what data is actually stored in the database. Types without prefix are the types meant to be used normally. Factored out some code to convert `ResponseStudent` to `Student` on the frontend.

- [x] implemented tracking survey timestamp per class on backend and frontend
- [x] lots of new types to more accurately show what's going on

### Test Plan 

Example response in Insomnia. Tested submitting the survey multiple times and confirmed different timestamps were recorded. Submitting the same class in a future survey does not affect the original timestamp. Also tested new types in the app to verify they work correctly.

Screen Shot 2022-07-09 at 19 31 01

### Notes 

Typed some things, but not everything. Mainly updated in the functions that I am most familiar with, there may be other places the types could be incorporated.

### Breaking Changes 

`submissionTime` is stored for each element of the student's `groups` array instead of in each student document at the top level.

When this PR is about to be merged I will write a script to convert the staging and production databases to match the expected structure.

### DIFF
--- a/backend/functions/src/student/functions.ts
+++ b/backend/functions/src/student/functions.ts
@@ -5,23 +5,26 @@ import {
   mapCatalogNameToCourseId,
   MissingCourseError,
 } from '../course/get_course_id'
-import { Student } from '../types'
+import { FirestoreStudent, Student } from '../types'
 const courseRef = db.collection('courses')
 const studentRef = db.collection('students')
 
 /** Get student data as Student type (with email and timestamps as Date) */
 export const getStudentData = async (email: string) => {
-  const studentData = (await studentRef.doc(email).get()).data()
-  if (!studentData) {
+  const studentDoc = await studentRef.doc(email).get()
+  if (!studentDoc.exists) {
     throw new Error(`Student document for ${email} does not exist`)
   }
-  studentData.email = email
-  studentData.groups = studentData.groups.map((group: any) => ({
-    ...group,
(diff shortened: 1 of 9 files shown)


backend/functions/src/student/functions.ts
```diff
@@ -142,6 +145,7 @@ export const addStudentSurveyResponse = async (
     groupNumber: -1,
     notes: '',
     notesModifyTime: surveyTimestamp, // Can't use serverTimestamp in arrays...
+    submissionTime: surveyTimestamp,
```
Coreball, line 148: This line of code implements the PR

wizhaaa (approved):
Look good

PR #85, opened 2022-07-18
Emailing Individuals Backend

CollinWoo:
### Summary 

This pull request modifies the `email/send` API endpoint to accommodate sending "no match" emails to individual students. This entails allowing requests to the endpoint to specify an individual recipient in lieu of a group, and updating the email send timestamp for an individual student after the email has been sent.

- [x] added functionality to send individual emails to `email/send` endpoint
- [x] added location in the Firestore database to log timestamps for emails sent to individual students
- [x] `email/send` endpoint now infers email recipients based on group number instead of having requests explicitly define the recipients

### Test Plan 

Test data was obtained by running `npm run populate.` Testing was conducted by submitting requests to email individual students using Insomnia, looking at backend logs, and verifying that the Firestore database was updated with the correct timestamps.

Sample request using Insomnia
Screen Shot 2022-07-18 at 12 25 21 PM

Backend logs verify emails were sent and timestamps were updated
Screen Shot 2022-07-18 at 12 26 34 PM

Timestamp for "no match" email updated in Firestore database
Screen Shot 2022-07-18 at 12 28 09 PM

### Notes 

- since updating timestamps occurs after an email is already sent, there is the possibility that an email will be sent to a student but the student's timestamp will not be updated. This should only occur in the rare case where an email is sent to an individual who does not belong to the course specified in the request.

### Breaking Changes 

- added a `firstNoMatchEmailTime` field to the groups array within the students collection in the Firestore database

### DIFF
--- a/backend/functions/src/emailing/functions.ts
+++ b/backend/functions/src/emailing/functions.ts
@@ -6,7 +6,7 @@ import { logger } from 'firebase-functions'
 import { db } from '../config'
 const courseRef = db.collection('courses')
 const studentRef = db.collection('students')
-import { FirestoreGroupMembership, FirestoreStudent } from '../types'
+import { FirestoreStudent } from '../types'
 
 // ==== Timestamp helper functions
 
--- a/backend/functions/src/emailing/functions.ts
+++ b/backend/functions/src/emailing/functions.ts
@@ -241,10 +241,13 @@ export const sendStudEmails = async (
   template = 'Share matched results',
   indivEmail?: string
 ) => {
-  const emailRcpts = await getRecipients(courseId, group, indivEmail)
+  let emailRcpts: string[]
+
+  emailRcpts = await getRecipients(courseId, group, indivEmail)
+
   const message = createEmailAsJson(emailRcpts, subject, body)
 
-  return await sendMails(
+  await sendMails(
     from,
     message,
     authToken,
@@ -253,4 +256,24 @@ export const sendStudEmails = async (
     template,
     indivEmail
   )
+    .then((result) => {
+      if (result === 202) {
+        logger.info(
+          `Email sent successfully by ${from} to ${emailRcpts.toString()}`
+        )
+      } else {
+        logger.error(
+          `** Likely auth error ** : Email failed to send by ${from} to ${emailRcpts.toString()}`
+        )
+
+        throw Error('Email send failure')
+      }
+    })
+    .catch((err) => {
+      logger.error(
+        `Email send request failed from ${from} to ${emailRcpts.toString()}`
+      )
+
+      throw err
+    })
 }
--- a/backend/functions/src/emailing/functions.ts
+++ b/backend/functions/src/emailing/functions.ts
@@ -73,12 +73,16 @@ const GRAPH_ENDPOINT = 'https://graph.microsoft.com'
 
 export async function getRecipients(
   courseId: string,
-  group: string,
-  indivEmail: string
+  group?: string,
+  indivEmail?: string
 ) {
   let emailRcpts = ['lscstudypartners@cornell.edu']
 
-  if (parseInt(group) > 0) {
+  if (group && indivEmail) {
+    throw new Error('Both group and indivEmail are specified')
+  }
+
+  if (group && parseInt(group) > 0) {
     const groupData = await courseRef
       .doc(courseId)
       .collection('groups')
@@ -88,7 +92,7 @@ export async function getRecipients(
     emailRcpts = [...emailRcpts, ...(groupData.data() as any).members]
   }
 
-  if (indivEmail !== undefined) {
+  if (indivEmail) {
     emailRcpts = [...emailRcpts, indivEmail]
   }
 
@@ -170,14 +174,14 @@ export const sendMails = async (
   template = 'Share matched results',
   indivEmail?: string
 ) => {
-  if (!group || (parseInt(group) < 0 && indivEmail === undefined)) {
+  if ((!group || parseInt(group) < 0) && !indivEmail) {
     logger.error(
       ` Invalid group ${group} and invalid email "${indivEmail}" for updating timestamps `
     )
-    return
+    throw new Error('No valid group or individual email')
   } else if (group && indivEmail) {
     logger.error(`group and individual email cannot both be specified `)
-    return
+    throw new Error('Both group and individual email are specified')
   }
 
   try {
@@ -207,7 +211,7 @@ export const sendMails = async (
           )
       }
 
-      if (typeof indivEmail !== 'undefined') {
+      if (indivEmail) {
         await updateIndivTimestamp(courseId, indivEmail)
           .then(() =>
             logger.info(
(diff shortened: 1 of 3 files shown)


backend/functions/src/emailing/functions.ts
```diff
@@ -133,18 +156,41 @@ export const sendMails = async (
...
+          .catch((err) =>
+            logger.error(
+              `Failed to update no match timestamp for course ${courseId} for student with email ${indivEmail}. Resulted in err: ${err.message} `
+            )
+          )
+      }
+
+      if (parseInt(group) < 0 && typeof indivEmail === undefined) {
+        logger.error(
+          ` Invalid group ${group} and invalid email ${indivEmail} for updating timestamps `
```
Coreball: I think this check should be earlier rather than later, so just in case the group and individual email are invalid it doesn't even try to send the emails before figuring out it can't update timestamps. Proactively detecting bad input is important for backend services and in this case we would want to error as early as possible. There should also be a guard for when there's both a group number and individual email supplied, because I think this an either-or scenario (unless that assumption is wrong, but at least it lines up with the frontend)?

backend/functions/src/emailing/routes.ts
```diff
@@ -24,20 +31,45 @@ const router = express()
...
+      .collection('groups')
+      .doc(group)
+      .get()
+
+    emailRcpts = [...emailRcpts, ...(groupData.data() as any).members]
+  }
+
+  if (indivEmail !== undefined) {
+    emailRcpts = [...emailRcpts, indivEmail]
+  }
```
Coreball: It might be worth putting this and the guard code in its own function so it would be easier to unit test if we later add backend testing. In other parts of the backend the callback functions given to `router.post` and similar are limited to only extracting data from the request `req` and passing it off to the main function, which may throw errors if it has problems or the input is invalid. The `.then` and `.catch` functions attached to the function call in this area of code handle sending the frontend the proper response, which has a success or failure HTTP error code attached.

backend/functions/src/emailing/routes.ts
```diff
@@ -1,8 +1,15 @@
 import express from 'express'
 import { logger } from 'firebase-functions'
-import { createEmailAsJson, sendMails, updateEmailTimestamp } from './functions'
+import { db } from '../config'
+import {
+  createEmailAsJson,
+  sendMails,
+  updateEmailTimestamp,
+  //updateIndivTimestamp,
```
Coreball: Nit: remove commented out code

backend/functions/src/emailing/functions.ts
```diff
@@ -5,6 +5,8 @@ import admin from 'firebase-admin'
 import { logger } from 'firebase-functions'
 import { db } from '../config'
 const courseRef = db.collection('courses')
+const studentRef = db.collection('students')
+import { FirestoreStudent } from '../types'
```
Coreball: Nit: imports should be before variable declarations

Coreball (commented):
Nice work! Overall very good job, great writeup and screenshots as well. It would also be good to explicitly call out somewhere that the frontend needs modification (next steps) to be fully compatible with the backend changes here, especially since the backend part here was written after the frontend was updated in #78. I tested sending a real email and the timestamps seem to be updating correctly working for me.

Regarding the situation in which updating the individual timestamp can fail, I don't think there's an easy way to fix this in all cases. One of the guards could be checking that the student is actually in the class before attempting any emailing, but this might be excessive. It's closer to a problem with our database consistency than malformed input. It's okay if this remains as a note, but thanks for pointing it out! (Or we could swap the order so it updates timestamps before emailing? Just a thought, but this doesn't work well because sometimes the email sending is _expected_ to fail like when the token is expired and the timestamps would be updated wrongly)

backend/functions/src/emailing/functions.ts
```diff
@@ -45,9 +48,53 @@ export const updateEmailTimestamp = (
...
+
+    emailRcpts = [...emailRcpts, ...(groupData.data() as any).members]
+  }
+
+  if (indivEmail !== undefined) {
+    emailRcpts = [...emailRcpts, indivEmail]
+  }
+
+  return emailRcpts
+}
```
Coreball: It feels weird to me that this function can still successfully get the recipients when the group and indivEmail combo is invalid. If both the group and individual email are defined it will return the individual email version.

Maybe create a function in `functions.ts` whose job is to get the recipients and send the email, and it does this by using `getRecipients` and `sendMails` (possibly `createEmailAsJson` too). Then the new function can be called in the `routes.ts` file to do everything, using these existing ones as helper functions.

backend/functions/src/emailing/functions.ts
```diff
@@ -119,9 +166,20 @@ export const sendMails = async (
...
 ) => {
+  if (parseInt(group) < 0 && indivEmail === undefined) {
+    logger.error(
+      ` Invalid group ${group} and invalid email "${indivEmail}" for updating timestamps `
+    )
+    return
+  } else if (parseInt(group) > 0 && indivEmail !== undefined) {
+    logger.error(`group and individual email cannot both be specified `)
+    return
+  }
```
Coreball: Specific to the validation logic, I think throwing errors would be better instead of returning void. It plays better with promises (the `.catch` on the promise will execute) and makes the return type of this function better defined, because now it might return void or a number, and it's not explicit that void is an expected return type or what it means (compared to an error).

Also may want to make the logic stronger to check that only one of group or indivEmail can be defined and are valid. Maybe make the group optional instead of default to -1, or something similar idk. Just something to better mirror the request JSON because now mentally we have to take into account that indivEmail will be the same as the request (a string or undefined) but the group number will always exist in this function, which may be the same as the request or maybe not (since it's default to -1).

Coreball (commented):
Good improvements! I'm looking at this during my workday so you can get feedback faster lol. There wasn't a place for me to add this as a comment on the code directly but always make sure to return a response to the frontend, otherwise I think Express might hang? This isn't a problem with your code, it's leftover from before but might cause you some issues if you add proper errors. The catch on sendEmails only logs an error to the logger but if you look at other routes.ts files it should also send an error code and a response.

backend/functions/src/emailing/routes.ts
```diff
@@ -24,38 +24,34 @@ const router = express()
...
+      courseId,
+      group,
+      template,
+      indivEmail
     )
+
+    res.status(200).json('Email send success')
+  } catch (err: any) {
+    res.status(400).json(err.message)
+  }
```
Coreball: Any reason why this was changed to a try-catch? Now it does not look like other routes.ts handlers

Coreball (approved):
Looks good! I tested the following cases:
- Both group number and individual email specified
- Invalid group number specified
- Invalid token

For other cases like when the course doesn't exist or the student's not in the course it still sends the email and logs an error to the console and this is totally OK.

backend/functions/src/emailing/routes.ts
CollinWoo: Mainly because the logging needed access to emailRecpts, which is now inside of sendStudEmails. I guess I could change it so sendStudEmails returns an object that contains the status code/list of recipients so that the .then code can be moved back to routes.ts.

Coreball: I just mean that this try-catch code is equivalent to
```js
sendStudEmails(...)
  .then(() => res.status(200).json('Email send success') // also i do not like how these are different than the rest of the backend where it's an object with a success field, but i don't know if that's even preferable
  .catch((err) => res.status(400).json(err.message)
```
which looks more like other routes.

Small thing but to me it's preferable if sendStudEmails is not abbreviated, there's not much benefit to the shorter version other than 3 characters and we're not trying to save those 3 letters. Any abbreviation has a small risk of being confusing in the future.

PR #87, opened 2022-08-05
Template editor add/update templates

Coreball:
### Summary 

This pull request adds the ability to create new templates and save changes to existing templates to the template editor. The new backend routes `/email/templates/add` and `/email/templates/update` are used on the frontend. When a new templates is created, the return value from `/email/templates/add` is the automatically-generated ID the document was created with in Firestore. This is also the name of the HTML file to be uploaded to Storage. The IDs are generated so LSC does not have to worry about thinking of a unique ID for each email and we do not need to do checks to confirm that IDs are unique. Buttons disable when their respective requests are happening.

- [x] implemented add new template on frontend
- [x] implemented update template/save changes on frontend
- [x] new backend route `POST /email/templates/add`
- [x] new backend route `POST /email/templates/update`

POST `/email/templates/add` to create a new template, returning `"data": "K694tDCbkpkS2o6Eo9xe"` for template ID:
```json
{
  "name": "New template",
  "type": "group",
  "subject": "Study partners!"
}
```

POST `/email/templates/update` to update an existing template, returns 400 Bad Request if there is no existing template with that ID
```json
{
  "id": "K694tDCbkpkS2o6Eo9xe",
  "name": "Updated template",
  "type": "student",
  "subject": "Study partners!?"
}
```

### Test Plan 

Visiting the template editor, the most recently edited template is shown, if there are none then it defaults to add template
Screen Shot 2022-08-05 at 00 15 42

Adding a new template
Screen Shot 2022-08-05 at 00 15 47

Filling out the template fields
Screen Shot 2022-08-05 at 00 17 07

After creating the template, it is automatically selected on the left and the buttons change to cancel and save changes
Screen Shot 2022-08-05 at 00 17 16

The update/save changes button works pretty much how you would expect. There is no check "are there changes to save" since there is no diffing, it simply overrides the existing fields. The template also goes to the top on the left since it's now the most recently modified template.

### Notes 

None

### Breaking Changes 

None

### DIFF
--- a/backend/functions/src/emailing/functions.ts
+++ b/backend/functions/src/emailing/functions.ts
@@ -312,7 +312,7 @@ const assertValidEmailType = (type: string) => {
 }
 
 /** Update an email template with new name, type, and subject */
-export const updateEmailTemplate = (
+export const updateEmailTemplate = async (
   id: string,
   name: string,
   type: 'group' | 'student',
@@ -346,7 +346,7 @@ export const updateEmailTemplate = (
  * Add a new email template with name, type, and subject.
  * @returns id of the newly created template
  */
-export const addEmailTemplate = (
+export const addEmailTemplate = async (
   name: string,
   type: 'group' | 'student',
   subject: string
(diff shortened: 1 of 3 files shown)


wizhaaa (approved):
Tried everything on my end and it worked as I expected. I also merged with my current branch  (#89) to see if it would all work together and it does as far as I have tested in terms of updating the list and displaying the correct template HTML body.

PR #86, opened 2022-07-28
Template editor view templates

Coreball:
### Summary 

This pull request is the first step towards implementing the email templates feature. Example email templates have been created in Firestore and Cloud Storage (the source and scripts are here). A new backend route was written to get template information. Frontend `/templates` page Template Editor makes a request to `/email/templates` on the back end API to load the template information from the `email_templates` collection in Firestore. Using that email templates information the frontend directly downloads the template HTML files from Storage. Then on the Template Editor page you can view the different templates and modify them locally, as well as cancel any changes. This PR does not implement saving changes.

- [x] figured out how email templates are stored in database
- [x] implemented load and view templates on Template Editor
- [x] implemented local editing of templates on Template Editor (HTML only)
- [x] implemented a cancel changes button on Template Editor (easy given how I designed other code)
- [x] set up Storage on the Firebase Emulator
- [x] add `add-example-templates.ts` script and data
- [x] new backend route `/email/templates` (no parameters)

### Test Plan 

https://user-images.githubusercontent.com/22627336/181440410-b4f6efb5-1076-4b7d-b277-22c68d0dcc39.mov

Video should capture most of the testing I did. Used the `add-example-templates.ts` script to add example templates. Also deployed and tested at https://zing-lsc.web.app/templates to make sure everything was working, had to fix CORS configuration for the storage bucket. Also had to change the storage bucket security rules.

```sh
$ export FIREBASE_STORAGE_EMULATOR_HOST=localhost:9199
$ export FIRESTORE_EMULATOR_HOST=localhost:8080
$ ts-node add-example-templates.ts
```

### Notes 

There's no entry point to the template editor yet. You have to go to the URL directly to access it.

Ran into CORS errors when deploying and testing on the development site. The download of the email templates is blocked by CORS. Fixed it by using the `gsutil` command line tool to change the bucket configuration, as explained https://firebase.google.com/docs/storage/web/download-files#cors_configuration

### Breaking Changes 

- New db collection `email_templates`
- New storage bucket `gs://zing-lsc-templates`

PR #89, opened 2022-08-08
Email edit new templates

wizhaaa:
# Summary 

This PR replaces the current hardcoded emailing templates that are selectable from the frontend email modal. 

- [x] Displays the email templates found in the `email_templates` collection. (run scripts in #86) 
- [x] Currently works and sends the 2 example templates HTML body. 
- [x] Also small style tweak of the group card when displaying a singular student card, displays it as grid-like. 

This PR is part of the frontend portion that continues off of #86 for custom email templates. 

# Test Plan 

1. Be sure to run the scripts in #86 to get the example templates 
2. Match students to groups 
3. Try emailing students and groups separately -> Should see a different list (1 item) based on if you selected to email students or groups. 

### With a group selected: 
image
image

### With a student selected: 
image
image

### Notes 

- Also changes the logic to check if the email was sent successfully. (Not sure how it managed to work before)
- No longer uses the old templates. Is possible to delete/remove the hard-coded templates and logic code in the future or reuse it somehow.

### DIFF
--- a/frontend/src/modules/EditZing/Components/EmailModal.tsx
+++ b/frontend/src/modules/EditZing/Components/EmailModal.tsx
@@ -335,58 +335,55 @@ export const EmailModal = ({
   }
 
   return (
-    <>
-      <ZingModal
-        open={isEmailing}
-        onClose={() => setIsEmailing(!isEmailing)}
-        containerWidth={'800px'}
-        containerHeight={'630px'}
-      >
-        <ZingModal.Title onClose={() => setIsEmailing(!isEmailing)}>
-          <Box
-            display={'flex'}
-            justifyContent={'center'}
-            textAlign={'center'}
-            paddingTop="20px"
-          >
-            <Typography variant="h4" component="h4" fontWeight={'700'}>
-              {title}
-            </Typography>
+    <ZingModal
+      open={isEmailing}
+      onClose={() => setIsEmailing(!isEmailing)}
+      containerWidth={'800px'}
+      containerHeight={'630px'}
+    >
+      <ZingModal.Title onClose={() => setIsEmailing(!isEmailing)}>
+        <Box
+          display={'flex'}
+          justifyContent={'center'}
+          textAlign={'center'}
+          paddingTop="20px"
+        >
+          <Typography variant="h4" component="h4" fontWeight={'700'}>
+            {title}
+          </Typography>
+        </Box>
+      </ZingModal.Title>
+      <ZingModal.Body>
+        {selectedTemplate ? (
+          <Box sx={{ padding: '1rem 3.5rem 0 3.5rem' }}>
+            {step <= 1 && <SelectTemplates />}
+            {step === 2 && <StepFailure />}
+            {step === 3 && <StepFinalFailure />}
           </Box>
-        </ZingModal.Title>
-        <ZingModal.Body>
-          {selectedTemplate ? (
-            <Box sx={{ padding: '1rem 3.5rem 0 3.5rem' }}>
-              {step <= 1 && <SelectTemplates />}
-              {step === 2 && <StepFailure />}
-              {step === 3 && <StepFinalFailure />}
-            </Box>
-          ) : (
-            <Box
-              sx={{
-                padding: '1rem 3.5rem 0 3.5rem',
-                textAlign: 'center',
-                display: 'flex',
-                flexFlow: 'column nowrap',
-                gap: '10rem',
-                alignItems: 'center',
-              }}
-            >
-              <Typography variant="h5">
-                No Templates for the selected group.
-              </Typography>
-              <Button href="/templates" sx={{ width: '250px' }}>
-                Add Templates
-              </Button>
-            </Box>
-          )}
-        </ZingModal.Body>
-        <ZingModal.Controls>
-          <BackButton />
-          <ProceedButton />
-        </ZingModal.Controls>
-      </ZingModal>
-      )
-    </>
+        ) : (
+          <Box
+            sx={{
+              padding: '1rem 3.5rem 0 3.5rem',
+              textAlign: 'center',
+              display: 'flex',
+              flexFlow: 'column nowrap',
+              gap: '10rem',
+              alignItems: 'center',
+            }}
+          >
+            <Typography variant="h5">
+              No Templates for the selected group.
+            </Typography>
+            <Button href="/templates" sx={{ width: '250px' }}>
+              Add Templates
+            </Button>
+          </Box>
+        )}
+      </ZingModal.Body>
+      <ZingModal.Controls>
+        <BackButton />
+        <ProceedButton />
+      </ZingModal.Controls>
+    </ZingModal>
   )
 }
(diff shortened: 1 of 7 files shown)


frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -23,10 +32,42 @@ export const EmailModal = ({
...
-    '' as TemplateName
+  // check if emailing students or groups
+  const recipientType = selectedStudents.length > 0 ? 'student' : 'group'
+
+  // template editor logic
+
+  const [templates, setTemplates] = useState<EmailTemplate[]>([])
+  const [templateName, setTemplateName] = useState('')
+  const [selectedTemplate, setSelectedTemplate] = useState<EmailTemplate>(
+    templates[0]
```
Coreball: This kind of removes some of the benefits of having the type system if we lie to it by giving it `templates[0]` (supposedly an EmailTemplate) knowing that `templates[0]` is actually undefined when this code runs.

frontend/src/modules/EditZing/Components/EmailPreview.tsx
```diff
@@ -39,7 +36,7 @@ export const EmailPreview = ({
       >
         <Box display={'flex'} flexDirection={'row'}>
           <Typography fontWeight={800}>Subject: </Typography>&nbsp;
-          <Typography>Study Partners!</Typography>
+          <Typography> {template.subject} </Typography>
```
Coreball: Nit: extraneous whitespace

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -23,10 +32,42 @@ export const EmailModal = ({
...
-  const [selectedTemplate, setSelectedTemplate] = useState<TemplateName>(
-    '' as TemplateName
+  // check if emailing students or groups
+  const recipientType = selectedStudents.length > 0 ? 'student' : 'group'
+
+  // template editor logic
+
+  const [templates, setTemplates] = useState<EmailTemplate[]>([])
+  const [templateName, setTemplateName] = useState('')
+  const [selectedTemplate, setSelectedTemplate] = useState<EmailTemplate>(
```
Coreball: Why are these separate? Doesn't one of these imply the other?

frontend/src/modules/EditZing/Components/EmailTemplateButtons.tsx
```diff
@@ -1,21 +1,31 @@
...
 
   const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    setSelectedTemplate(event.target.value as TemplateName)
+    const choice = event.target.value
+    templates.forEach((template) => {
+      if (template.name === choice) {
+        setSelectedTemplate(template)
+      }
+    })
+    setTemplateName(event.target.value)
```
Coreball: The reason why keying things on data like names instead of ids is generally not preferable is because there's no guarantee the template names are unique, unlike ids. But I can see how that would be difficult in this case because the RadioButtons component is really old and doesn't distinguish between values and labels.

Coreball (commented):
Great work! Seeing these custom templates on the email modal is really nice in watching the email templates make their way through development. There is a bug where on the email template selection screen, the next button is enabled and you can preview the email without choosing a radio button or seeing a radio button selected by default on the template selection page.

Hope to see this merged soon, good job!

frontend/src/modules/EditZing/Components/EmailTemplateButtons.tsx
wizhaaa: Yeah, I can see how this is actually a problem but I don't see how I could change the selected template any other way unless I rebuild the radio buttons/use the default radio buttons (not in our core lib) that can take in the id (to handle change) and name (to display). Is there any way to fix it as it is now to check id against the selected radio button?

frontend/src/modules/EditZing/Components/EmailModal.tsx
wizhaaa: That is what I was hoping I think since as you pointed out, selectedTemplate is actually undefined and the radio buttons don't actually work as intended when I try to implement using `selectedTemplate` and `selectedTemplate.name` (for the `currentValue` props for the buttons). I think the main issue is still the fact that selectedTemplate is undefined. I know you said to initialize it at the beginning of the array and how in the old implementation it was initialized to an empty string. 

What would I initialize selectedTemplate to be? I don't think an empty object `{}` works (doesn't type check). And if I don't initialize it, it won't allow me to pass in a possibly undefined object into the other components. 

Maybe one solution is to use the id and initialize the id as you had in template editor and then break down the template into subject + HTML which can be initialized into empty strings.

Coreball: > it won't allow me to pass in a possibly undefined object into the other components

What about changing those components to take a possibly undefined object? I think it makes sense for EmailTemplateButtons, as it means "there might be a selected template, or there might not". Then EmailTemplateButtons can determine how to handle that when it shows the radio buttons.

For the email preview, you can use `selectedTemplate!` since by the time the email is being previewed, they must have selected a template, if our app is working correctly. Then you don't need to change how EmailPreview works, it can still take in the email template and not worry about it being undefined.

frontend/src/modules/EditZing/Components/EmailTemplateButtons.tsx
Coreball: I think modifying `RadioButtons.tsx` would not be too hard, as the relevant existing code is just the following. You could just add another prop called `labels` and associate the values to labels by their index.

```jsx
<RadioButton
  value={value}
  label={value}
/>
```

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -292,39 +335,58 @@ export const EmailModal = ({
...
-      </ZingModal.Controls>
-    </ZingModal>
+        </ZingModal.Body>
+        <ZingModal.Controls>
+          <BackButton />
+          <ProceedButton />
+        </ZingModal.Controls>
+      </ZingModal>
+      )
+    </>
```
Coreball: Remove this extra parentheses and `<></>` tag

frontend/src/modules/EditZing/Components/EmailPreview.tsx
```diff
@@ -39,7 +35,7 @@ export const EmailPreview = ({
...
```
Coreball: Previous comment about extraneous whitespace still remains unfixed

frontend/src/modules/EditZing/Components/EmailTemplateButtons.tsx
```diff
@@ -1,21 +1,24 @@
...
 
+  // changing the selected email template
   const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    setSelectedTemplate(event.target.value as TemplateName)
+    const newSelection = event.target.value
+    templates.forEach((template) => {
+      if (template.id === newSelection) {
+        setSelectedTemplate(template)
+      }
+    })
```
Coreball: Cleaner with `setSelectedTemplate(templates.find(template => template.id === newSelection)!)`

Coreball (approved):
Good work! Tweak a few things and then feel free to merge this.

PR #90, opened 2022-08-08
Centralized application state course/student routes and React Context

Coreball:
### Summary 

This pull request is the first step moving towards a centralized application state model, where the app loads all course and student information at once. This will reduce delays waiting for information to be loaded on demand and unblocks CSV generation and certain sorting/filtering options on the dashboard that are required for Fall 2022 launch.

The updated GET `/course` route returns a list of all courses. It now returns group information for those courses as well, with the students referenced by their emails. The new GET `/student` route returns all students. Together these reduce the total information that would need to be transferred in some cases because if a student is in multiple courses, their student information exists once and is referenced by student email, instead of being transferred once per each of the multiple courses they are in.

New types and such are done with the existing pre-email-template timestamp format. These will be updated later. Moved the Student type definitions from being in the EditZing module to the frontend Core module.

For the Dashboard, Groups was renamed to CourseGrid and GroupCard was renamed to CourseCard to more accurately reflect what they are now.

For how the improvements from this PR can be used in practice, see Dashboard.tsx and how instead of needing to use useEffect and a spinner to load the courses, `const { courses } = useCourseValue()` can be used instead.

- [x] reworked backend route GET `/course`
- [x] new backend route GET `/student`
- [x] CourseContext and StudentContext (separate to give the update logic space later)
- [x] Dashboard uses centralized application course list
- [x] renamed some things

### Test Plan 

Frontend works the same as before, except there is no loading on the dashboard anymore.

Sample GET `/course`

Screen Shot 2022-08-07 at 19 27 58

Sample GET `/student`

Screen Shot 2022-08-07 at 22 54 27

### Notes 

Not included: query parameters to only get a list of courses for a specified semester. Maybe this will be added later but it's not necessary right now, as filtering can be done on the frontend (and this allows for instant semester switching).

### Breaking Changes 

- GET `/course` is not the same as before
- This migration is being done incrementally so everything else should still work fine


wizhaaa (approved):
Tested some functionalities on my end and everything is working well. Great work!

PR #92, opened 2022-08-12
Dashboard Select Semester

wizhaaa:
# Summary 

This PR allows the ability to select/filter the courses shown on the dashboard by roster. 

Accessiable by clicking name -> Select Semester -> Choose from submenu. 

# Test Plan 
- Populate and change some courses roster to something other than current (ie. 'SU22')
- Change sort order. Should sort the current roster. 
### Selected Summer
image

### Back to Fall 
image

## Notes 

- potentially migrate to side bar in future
- box shadow won't change sadge


frontend/src/App.tsx
```diff
@@ -54,6 +54,7 @@ const App = () => {
         user
           .getIdToken(true)
           .then((idToken) => {
+            console.log('idToken', idToken)
```
Coreball: Is this meant to be part of the PR

frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -134,7 +145,30 @@ export const Dashboard = () => {
...
+  }
+
+  const sortedCourses = sorted(currentCourses, sortedOrder)
+
+  useEffect(() => {
+    const currentSemesterCourses = courses.filter(
+      (course) => course.roster === 'FA22'
+    )
+    setCurrentCourses(currentSemesterCourses)
+  }, [courses])
```
Coreball: An alternative for this would have been to do something like
```js
const [selectedRoster, setSelectedRoster] = useState('FA22')
const sortedCourses = sorted(
  courses.filter(course => course.roster === selectedRoster),
  sortedOrder
)

<MenuItem onClick={() => setSelectedRoster('SU22')}>
```
Curious to hear your thoughts on this. Both ways are definitely acceptable, but this is what I thought of first (it's more similar to how the sorting works with only the selection being kept in state). Yours may be slightly more efficient in that the courses are only filtered on demand.

But I guess if `courses` changes then the user's semester selection will be thrown out since the useEffect will run again. The value of courses might change like when the user modifies courses on the edit page since that state is shared. So it's possible this could cause issues, but I haven't tested and that code hasn't been written yet.

Storing the selected roster as a state variable may make filtering the CSV based on roster more straightforward.

Coreball (approved):
Nice work! It behaves as expected and using the submenu successfully changes the current semester that's shown on the dashboard. Since everything works I would be fine with merging this in now, but I would also like to suggest some changes. IMO, it makes more sense to me that the more recent semesters would be at the top, but I'm willing to hear what other people think.

I haven't thought too much about the sidebar. Hard coding the semester menu items and default roster is fine since we wouldn't have to update them for a while. We already have to manually edit what the default semester for survey submissions is. Agree that the menu shadows are a bit harsh, but they don't cause me too much trouble personally. Haven't looked into fixing them. It would make sense to show the current semester on the dashboard, but is not critical.

I made some changes to this PR. Removed the need for the outer `Button` wrapping the `MenuItem` by changing the event type of the click handler. Changed the left arrow to be a more general `ChevronLeftIcon` instead of `ArrowBackIosIcon` and gave it theme-defined color. Personally I like the slightly smaller one. Fixed the anchoring of the submenu to be what we want, with the top right at the top left of the menu item that opens it.

frontend/src/App.tsx
wizhaaa: nope i forgor to remove lol

frontend/src/modules/Dashboard/Components/Dashboard.tsx
wizhaaa: Changed so the buttons change the selected roster instead, which also makes the selected roster information available to the rest of the dashboard (for things like CSV). I think I didn't expect the rosters to change when being stored in const and not a state, but I'm assuming it updates since it's based on two other states.

I wasn't exactly sure about the best way to do this either as I know they may want a filtering system in the future. But I'd say this new solution works fine and should still be easy to build the filtering logic on top of it.

PR #94, opened 2022-08-13
Template editor timestamps and tooltips

CollinWoo:
### Summary 

This pull request implements a new representation for email timestamps in the database to accommodate custom templates and adds tooltips to display the timestamps on the frontend.

- [x] implemented new timestamp representation of type {[key string]: timestamp}
- [x]  altered existing code to accommodate the change
- [x] implemented showing tooltips for individuals and groups on the course page

### Test Plan 

Testing was done by populating the database with sample students and templates using `npm run populate` and `ts-node add-example-templates.ts` and sending email requests with Insomnia.

Individual timestamps
Screen Shot 2022-08-12 at 3 32 12 PM

Group Timestamps
Screen Shot 2022-08-12 at 3 32 19 PM

Unmatched Timestamps
Screen Shot 2022-08-12 at 3 34 18 PM

### Notes 
- There's a lot of converting between objects and arrays in the code (mainly for mapping keys) that makes it messy. This could be negated by either changing the timestamp representation in the database (or response from the backend) from an object to an array (which can be mapped over directly), or by having an object map function that can be accessed from any file
- Occasionally when I was testing the code, there would randomly be an error when accessing a course page. The error would go away after a refresh and the page would load normally afterwards. This probably has something to do with concurrency, but I haven't been able to reproduce it on later commits so either the problem went away by itself or it is very rare
- the tooltips are displayed left to right in alphabetical order by name

### Breaking Changes 

- Email template timestamps changed from hardcoded fields to an object of Template id:timestamp key/value pairs


backend/functions/src/emailing/functions.ts
```diff
@@ -218,33 +238,53 @@ export const sendMails = async (
...
-              `Failed to update no match timestamp for student with email ${indivEmail} for course ${courseId}. Resulted in err: ${err.message} `
+              `Failed to update ${template} timestamp for student with email ${indivEmail} for course ${courseId}. Resulted in err: ${err.message} `
             )
           )
       }
     }
     return response.status
   } catch (error) {
-    return error
+    throw error
```
Coreball: VSCode tells me that you could just remove the whole try/catch part if the error is just being thrown without any other actions

backend/functions/src/matching/functions.ts
```diff
@@ -118,9 +115,6 @@ async function makeMatches(courseId: string) {
       memberData: newGroup,
       createTime: new Date(),
       updateTime: new Date(),
-      shareMatchEmailTimestamp: null,
-      checkInEmailTimestamp: null,
-      addStudentEmailTimestamp: null,
```
Coreball, line 123: I think you are missing `templateTimestamps` because no empty timestamps object for groups is returned in the response from `/make`

Actually, I see that you just handle the empty timestamps object on the frontend. That works too

backend/functions/src/types/index.ts
```diff
@@ -24,9 +24,7 @@ export type Group = {
   createTime: Date
   updateTime: Date
   // TODO replace the below with the new template timestamps
-  addStudentEmailTimestamp: Date | null
-  checkInEmailTimestamp: Date | null
-  shareMatchEmailTimestamp: Date | null
+  templateTimestamps: { [key: string]: Date }
```
Coreball, line 26: Remember to remove the TODO comment above, otherwise might be confusing in future (other files too)

frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -74,9 +84,7 @@ export const Dashboard = () => {
                 membership.groupNumber !== -1
                   ? `${course.names.join('/')}_${membership.groupNumber}`
                   : undefined,
-              dateShareMatchEmail: group?.shareMatchEmailTimestamp?.toLocaleString(),
-              dateCheckInEmail: group?.checkInEmailTimestamp?.toLocaleString(),
-              dateAddStudentEmail: group?.addStudentEmailTimestamp?.toLocaleString(),
+              dateTemplateEmails: localeMap(group?.templateTimestamps),
```
Coreball: Don't think this is what they want

Screen Shot 2022-08-13 at 02 06 53

frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -87,6 +92,26 @@ export const EditZing = () => {
     }
   }
 
+  //Map for getting the names of templates based on ID for rendering tooltips
+  const [templateNameMap, setTemplateNameMap] = useState<{
+    [key: string]: string
+  }>({})
```
Coreball, line 101: This should be simpler once we centralize the templates across the app state later

frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -312,7 +318,9 @@ export const EditZing = () => {
   const handleSelectNewlyMatched = () => {
     setSelectedGroupNumbers(
       studentGroups
-        .filter((group) => group.shareMatchEmailTimestamp === null)
+        .filter(
+          (group) => group.templateTimestamps['share-match-email'] == null
+        )
```
Coreball, line 326: Worth noting that this is hardcoded at the moment

frontend/src/modules/EditZing/Types/CourseInfo.ts
```diff
@@ -7,24 +7,25 @@ export interface CourseInfo {
   unmatched: string[]
 }
 
+export interface Timestamp {
+  template: string
+  timestamp: Date
+}
```
Coreball: Is this used?

frontend/src/modules/EditZing/Types/CourseInfo.ts
```diff
@@ -39,3 +40,15 @@ export interface CourseStudentDataResponse {
...
+  success: boolean
+  data: {
+    id: string
+    name: string
+    type: 'group' | 'student'
+    subject: string
+    body: string
+    modifyTime: Date
+  }[]
+}
```
Coreball: Is this the same as `EmailTemplatesResponse` in `@core/Types`? I think this whole file may be deprecated soon in favor of centralizing shared types across different components.

Coreball (commented):
Very good job Colin! This pull request was really nice to read, the code overall looks really good and it's well documented and easy to understand. I don't have too much to say about it as a result, but you ought to fix the CSV output before we merge this in. Really great job!

frontend/src/modules/Emailing/Components/Emailing.tsx
```diff
@@ -161,7 +161,7 @@ export const sendEmail = async (emailItems: any) => {
       emailAddress: email,
       emailBody: emailBody,
       emailSubject: emailSubject,
-      emailRcpts: emailRcpts,
+      indivEmail: emailRcpts,
```
Coreball: Would be clearer if it was just called `indivEmail` everywhere

Coreball (commented):
Thanks for making the improvements. Have you tested end to end yet? You should have access to the LSC inbox now to test for real. During my testing, sending emails showed the error message on the frontend every time even though the email send requests were actually succeeding. I think the error handling needs to be refined since it isn't really in a production-ready state yet.

The CSV columns for specific email timestamps are also missing if no one has been sent that email yet. I think this is worth fixing at some point but you can just make it a note when we ship it that the columns won't appear unless it has been sent at least once. This is because without a the key being defined in one of the objects that turns into the CSV, it doesn't appear as a header.

Coreball (approved):
Nice work Colin!

PR #95, opened 2022-08-16
Home Page Separation 

wizhaaa:
# Summary 

Separates the website's root page to only allow access to the student survey and not the LSC Admin Login/Dashboard. 

Styling was based on LSC feedback. 

- [x] also adds the back button on an edit zing page. The Zing groups icon is now also an icon button that goes to the dashboard. 
- [x] new admin home page 

# Test Plan 

### Root / 
image

### Admin page /admin
image

### Breaking Changes 
- Admin + LSC staff now either need to go "/dashboard" or "/admin" to login and access their dashboard.


frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -392,7 +393,17 @@ export const EditZing = () => {
           borderBottomColor: 'essentials.12',
         }}
       >
-        <LscIcon sx={{ height: '50px', width: '50px' }} />
+        <IconButton
+          color="secondary"
+          href="/dashboard"
```
Coreball: Using `href` results in reloading the whole web app when it goes to the other page, you can use `<IconButton component={Link} to="/dashboard">` instead. That uses the Link component from React Router. Not sure if we actually use this yet in our codebase.

frontend/src/modules/Home/Components/Home.tsx
```diff
@@ -1,59 +1,97 @@
...
+          }}
+        >
+          <Box
+            sx={{
+              width: '100%',
+              maxWidth: '50px',
+            }}
+          >
+            <CornellSeal />
+          </Box>
```
Coreball: I believe you are (officially) supposed to use the other logo if it's smaller than 73px, according to https://brand.cornell.edu/logos/academic/

frontend/src/modules/AdminHome/Components/AdminHome.tsx
```diff
@@ -0,0 +1,97 @@
...
+              width: '100%',
+            }}
+          >
+            Cornell Learning Strategies Center
+          </Typography>
+        </Box>
+        <Typography
+          variant="h2"
+          fontWeight={'600'}
+          sx={{ mb: '3rem', color: 'white', fontSize: '3.25rem' }}
```
Coreball: Why is the font size different than the regular home page even though they say the same thing

frontend/src/modules/AdminHome/Styles/Home.style.tsx
```diff
@@ -0,0 +1,31 @@
+import styled, { css } from 'styled-components'
```
Coreball: Should this file be called `AdminHome.style.tsx`?

frontend/src/modules/AdminHome/Styles/Home.style.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export const StyledLeftPanel = styled.div`
+  ${panel};
+  align-items: left;
+  margin-left: 5rem;
+`
+
+export const StyledRightPanel = styled.div`
+  ${panel};
+`
```
Coreball: You can delete these if they are not being used (in the regular Home.style.tsx too)

Coreball (commented):
Good, but should the main landing page at `/` now have its button be a primary button since it's the only action students can take there? I do question the purpose of the page a bit now because it's just a gateway for students to go to the survey, but that's a question for later.

The font size difference between the home and admin home pages is a bit strange. Also when going to `/dashboard` I am redirected to `/` as normal. I feel like this and other private routes should redirect to `/admin` so they have a way to log in without going to the `/admin` URL manually if their session expires.

frontend/src/modules/Home/Components/Home.tsx
wizhaaa: Just changed to 73px. I saw this too and originally tried using that but I wasn't able to crop the image. As, I could convert to JPG and then crop but its just a white block since JPG doesn't remove background. Might be possible in a photoshop tool so perhaps Prachi/designer could work on that and replace it if need be.

frontend/src/modules/AdminHome/Components/AdminHome.tsx
wizhaaa: Yeah good catch I was trying to fiddle with the sizing and didn't make sure they were the same.

wizhaaa:
I do agree with you about the point of the home screen since in practice we only send students the survey link so they shouldn't hit this page. But, then there is the possibility of them somehow ending up here but the current home page does seem a bit weird to me but could be something to update with a solid idea of what it should be.

Coreball (approved):

PR #96, opened 2022-08-19
Starter email templates for Fall 2022

Coreball:
### Summary 

Adds the starter email templates for Fall 2022. These can be customized during the middle of the semester without Zing team intervention by using the template editor at `/template`. Also exports the email templates to the emulator so when running `npm run emulate`, the email templates will already be in the database and storage emulator. No need to run `ts-node add-example-templates.ts` every time we test the backend during development.

- [x] script to create starter email templates for Fall 2022
- [x] starter email templates are part of the emulator

### Test Plan 

Template editor viewing the template for sharing matches:

Screen Shot 2022-08-19 at 01 05 02

### Notes 

All templates have TEST prepended to their email subjects when adding them via the script or running the emulator. This is an additional safeguard to make it clear what emails are for testing. For production the script will be run once, then the templates will be modified to remove the TEST. This is important because both test emails and real emails are sent from the same inbox.

We don't have value substitution yet, so for individual emails we are not able to dynamically change the email depending on which course it is being sent for.

### Breaking Changes 

- Hard coded email template IDs as `share-match`, `no-match-yet`, `add-student`, and `check-in`.


wizhaaa (approved):
The HTML looks good so far to me.

PR #97, opened 2022-08-23
Implement Filtering by Unsent Check-In Email

CollinWoo:
### Summary 

This pull request implements a filtering option that only displays courses where not all groups have recieved check-in emails yet.

- [x] implemented filtering by unreceived check-in email
- [x] added new filtering option to "sort by" dropdown 

### Test Plan 

Testing was done by populating the firestore database using `npm run populate,` making matches, and sending check-in emails to some groups in certain courses but not others, and then verifying that the correct courses show up when filtering.

Screen Shot 2022-08-23 at 4 30 11 PM

### DIFF
--- a/frontend/src/modules/Dashboard/Components/Dashboard.tsx
+++ b/frontend/src/modules/Dashboard/Components/Dashboard.tsx
@@ -27,6 +27,7 @@ type SortOrder =
   | 'matchable-first'
   | 'classes-a-z'
   | 'classes-z-a'
+  | 'no-check-in-email'
 
 export const Dashboard = () => {
   const { user } = useAuthValue()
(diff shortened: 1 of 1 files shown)


frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -133,7 +139,7 @@ export const Dashboard = () => {
           } else return 1
         })
       case 'classes-a-z':
-        return [...courseInfo].sort((a, b) => {
+        return courseInfo.sort((a, b) => {
```
Coreball: The `[...courseInfo]` copies the array to avoid mutating the state variable since `.sort()` is in-place in JavaScript. You should probably keep it

Coreball (approved):
Good, glad the implementation was so simple after #90 . Fix an unneeded change and then merge this in because the other filter is more important, according to our meeting notes. Or modify this PR to include both because there's not many lines of code for each.

PR #98, opened 2022-08-25
implement filter for courses with unsent no match emails

CollinWoo:
### Summary
This pull request implements a filtering option that only displays courses where not all unmatched students have recieved a no match email yet.

 - [x] implemented filtering by unreceived no-match email
 - [x] added new filtering option to "sort by" dropdown

### Test Plan
Testing was done by populating the firestore database using npm run populate, making matches, and sending no-match emails to some unmatched students in certain courses but not others, and then verifying that the correct courses show up when filtering.

Screen Shot 2022-08-25 at 12 05 12 AM

### DIFF
--- a/frontend/src/modules/Dashboard/Components/Dashboard.tsx
+++ b/frontend/src/modules/Dashboard/Components/Dashboard.tsx
@@ -28,6 +28,7 @@ type SortOrder =
   | 'classes-a-z'
   | 'classes-z-a'
   | 'no-check-in-email'
+  | 'no-match-email'
 
 export const Dashboard = () => {
   const { user } = useAuthValue()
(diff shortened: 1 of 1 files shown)


frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -28,6 +28,7 @@ type SortOrder =
   | 'classes-a-z'
   | 'classes-z-a'
   | 'no-check-in-email'
+  | 'no-match-email'
```
Coreball: The wording of this is a little inconsistent with the previous sort order type for check in emails. That was (no) check in email and this is (no) match email, except it's like (no) no match yet email

Coreball (approved):
Good job

Coreball:
@CollinWoo I think this is not working in production right now because for the naming convention
`<MenuItem value="no-match-email">`
is supposed to be
`<MenuItem value="no-no-match-email">`
to match with the `SortOrder` type. Please make a small fix for this soon!

Coreball:
Will be fixed in #100

PR #99, opened 2022-08-31
Value substitution for No Match Yet

Coreball:
### Summary 

This pull request adds basic value substitution, enabling usage of `{{COURSE_NAME}}` in the HTML template. The template editor will replace `{{COURSE_NAME}}` with `ABC 1100`, and the email modal will replace it with the actual name of the course the no match yet emails are being sent for, such as `CS 1110`. For cross-listed courses they will be joined with a slash, like `CS 2110/ENGRD 2110`. The value substitution is also present in the sent email.

Also added is an option to select all unmatched students for which the no match yet email has not been sent yet. This is accessed through the "Send email to" button drop down menu. Had to change where the StudentCard selection is from being handled in the card's state to instead depend on selectedStudents in the main component.

There is a table on the template editor to see what substitutions are available. They are handled with a map to support other substitutions, though getting the values for those substitutions may be the tricky part. Emailing each student something different (like their name) in the same email modal would require some reworking because the same template replacement is applied to all requests. How would it be displayed on the preview too? Also updated the emulator to use the newer version of the no-match-yet email with the value sub.

- [x] implemented basic value substitution for `{{COURSE_NAME}}`
- [x] updated email template "No Match Yet" and saved in the emulator
- [x] "Send email to" drop down now has an option for no match yet

### Test Plan 

Value substitution preview in the Template Editor. Table in the top right with available substitutions is new too

https://user-images.githubusercontent.com/22627336/187570386-cc63b379-70c8-4a30-96ed-17409bce1ff3.mov

Value substitution on the Email Modal + "Send email to" drop down only selects unmatched students

https://user-images.githubusercontent.com/22627336/187570506-8668ae7e-20e4-4e80-9da3-c0e80e1f6e01.mov

Sent email example

Screen Shot 2022-08-30 at 21 07 12

### Notes 

Student needs the no match yet email sent separately for each course where they do not have a match yet.

### Breaking Changes 

None


CollinWoo (approved):
Everything works perfectly on my end! The code is very concise and readable as well.

PR #100, opened 2022-09-02
Fix unsent no match email bug

CollinWoo:
### Summary 

This pull request fixes a bug for filtering classes by unsent no match emails

- [x] changed the value of a Menu-Item from`no-match-email` to `no-no-match-email`
- [x] that's it 

### Test Plan
Testing was done by populating the firestore database using npm run populate, making matches, and sending no-match emails to some unmatched students in certain courses but not others, and then verifying that the correct courses show up when filtering.


Coreball (approved):
Worked in my testing

PR #102, opened 2022-09-23
Centralized application state for Edit Page

Coreball:
### Summary 

This pull request reworks the edit page (`/edit/FA22-358526`) to use the centralized application state introduced in #90. The edit page now reads from the shared courses and student data and updates it when modifications happen, such as moving students between groups, adding email timestamps, matching students/forming groups, and updating student notes.

An application-wide loading screen for private routes has also been added to make sure course and student data has been loaded before rendering the dashboard or edit page. Some backend routes were updated or modified to be compatible, namely the `/make` endpoint only returns student email addresses.

This PR makes viewing the edit page happen instantly instead of needing to sit through a loading screen. This pull request was getting pretty large so I split it up into multiple Notion tasks, but this first PR is still quite large.

- [x] add application-wide loading screen for private routes
- [x] change edit page to render from shared courses and students
- [x] centralize update student notes
- [x] fix modal hiding before note save finished
- [x] centralize moving students between groups or unmatched
- [x] centralize make groups (match students)
- [x] change backend route for `/make` to return emails and not student objects
- [x] change frontend add email timestamp to not call api
- [x] centralize add email timestamps
- [x] fix student document group membership consistency bug for moving students back to unmatched (did not reset to -1)

### Test Plan 

Application-wide loading screen for private routes (happens once when loading the app initially) (simulated load time delay)

https://user-images.githubusercontent.com/22627336/191890234-e703ff90-f0bc-4191-b976-3810c7ee9a82.mov

Centralized application state removes loading screens between every edit page, making editing instant!

https://user-images.githubusercontent.com/22627336/191890334-667a22d3-8557-4baa-9d51-fc6a85406ae9.mov

Before this PR the edit page had to load every time (before #90 the dashboard also reloaded every time you went back to it)

https://user-images.githubusercontent.com/22627336/191890457-93828920-076f-41df-aacd-bf43c55c8af3.mov

### Breaking Changes 

- `/make` returns different data now, but this is the only place where it's used


CollinWoo (approved):
Great stuff! For such a large PR, the code is very well-written and readable, and the instant editing is impressive to see. I like how this PR simplifies things such as variable names and data from the backend in addition to implementing such an integral feature. I'm implementing a similar feature for the template editor, so having such a well formatted model to work off of is super helpful!

PR #104, opened 2022-10-03
Updated emails for populating emulator

mluo24:
### Summary 

Pull request adds @rgu0114 and @JustKong13 to the email lists so that they can log in via the emulator, in addition to the new LSC member. Also, the import data has been updated with `firebase emulators:export /emulator` where `/emulator` is the absolute path to where it was originally in the file tree.

### Test Plan 

Ran started backend emulator and ran `npm run populate` to test out to see the emails showed up in the random classes as well as the Firestore database. Also checked to see if the `allowed_users` are added properly on starting the emulator.
image


JustKong13 (approved):
Yayyy!!!! now we can develop without having to restart everything everytime!!

PR #106, opened 2022-10-06
Fix group ordering

JustKong13:
### Summary 

Fixed the bug where the order of number of the group in a class page would show up in alphabetical order rather than numerical order

### Breaking Changes 


frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -38,6 +38,7 @@ import { KeyboardArrowDown, KeyboardArrowUp } from '@mui/icons-material'
 import { useAuthValue } from '@auth/AuthContext'
 
 import { DASHBOARD_PATH } from '@core/Constants'
+import { forEachChild } from 'typescript'
```
b3njamint: What's this for—is it needed?

b3njamint (commented):
Good job finding the issue! In the future, it would be nice to see some screenshots of your fix in the test plan. Thanks for removing the commented code too :)

Also, check my comment!

frontend/src/modules/EditZing/Components/EditZing.tsx
JustKong13: no its not i think it auto added :/

Coreball (commented):
Good work, but I believe that the current implementation of the group ordering fix does not accurately resolve the sort order, since it relies on changing the labels that are displayed for each group instead.

There is a difference between the index and the `groupNumber` property. We want to show the correct group number for each group, not relabel which group is which based on the array order (this is why `groupNumber` exists).

Here is one instance where this may cause us trouble. Say there are groups 1 to 9 present, showing up as `1, 2, 3, ..., 9` for the order of the group cards. Then group 10 is created. The way the data is returned from the backend by default will cause the group order to go `1, 10, 2, ..., 9`. Displaying the index instead of the group number will make it _look_ like `1, 2, 3, ..., 10`, but "group 2" is actually group 10 with a different name! This would be confusing because our users would expect that people stay in the same groups. In the database it's still be keyed with the original group number and that wouldn't match up with what's shown on the frontend.

JustKong13:
> Good work, but I believe that the current implementation of the group ordering fix does not accurately resolve the sort order, since it relies on changing the labels that are displayed for each group instead.
> 
> There is a difference between the index and the `groupNumber` property. We want to show the correct group number for each group, not relabel which group is which based on the array order (this is why `groupNumber` exists).
> 
> Here is one instance where this may cause us trouble. Say there are groups 1 to 9 present, showing up as `1, 2, 3, ..., 9` for the order of the group cards. Then group 10 is created. The way the data is returned from the backend by default will cause the group order to go `1, 10, 2, ..., 9`. Displaying the index instead of the group number will make it _look_ like `1, 2, 3, ..., 10`, but "group 2" is actually group 10 with a different name! This would be confusing because our users would expect that people stay in the same groups. In the database it's still be keyed with the original group number and that wouldn't match up with what's shown on the frontend.

Ohhh, okay, that makes sense, thanks for the catch!

Coreball (commented):
Hey Justin, I'd suggest merging `main` into your branch because you'll find that the place where you added the new code has since been deleted in #102 as part of the migration to keep all data shared and consistent across the whole app.

Adding onto what @b3njamint said, it would be good to see screenshots of the fix in action in the PR body under "Test Plan," as it visually shows to everyone that the fixes work. You should also describe how you tested it too, e.g. there were 9 groups, created a 10th group and verified it showed up at the end, refreshed the page and it was still at the end.

I would also recommend looking into [Array.sort()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort), specifically its usage with a compare function, to do the sorting. It is cleaner to read than a for loop and much more preferable. There are examples of it in use on the Dashboard, like this snippet that sorts the course cards by which course most recently got requests: https://github.com/cornell-dti/zing-lsc/blob/d645fdfdf944136d98a83f4ae51ddb36c02a98f9/frontend/src/modules/Dashboard/Components/Dashboard.tsx#L134-L138

(the `[...courseInfo]` was to ensure we copy the array before sorting, since Array.sort() mutates the array and this function was originally operating directly on the state variable from `useState`, which shouldn't be modified without calling the `setState`).

JustKong13:
New PR was made

PR #108, opened 2022-10-13
Add Firebase logging for API requests

rgu0114:
### Summary of Changes 

Added basic logging for all API requests 

- for requests that require authorization, we log user uid and display name in Firebase 
- for all requests, added a new middleware function that logs request params and request body
Screen Shot 2022-10-20 at 7 46 29 PM

Screen Shot 2022-10-20 at 7 36 02 PM

10/20 Changes 

- simplified everything by outputting stuff using JSON 
- logged request type as well

10/13 Changes 

- added request endpoint, params, and body to a single log output
- trying to format it nicely but \n for newline doesn't work for some reason
### Testing 
Caller info shows up in the log when we first log into Zing LSC:
Screen Shot 2022-10-13 at 3 35 38 PM

Logging still works when we submit a form and match students for a course:
Screen Shot 2022-10-13 at 3 36 53 PM

Mostly outputs everything as expected but new lines aren't working:
Screen Shot 2022-10-13 at 8 23 21 PM

### Todo 
Request params are currently showing up as empty 
Find a way to also log which endpoint(s) are being hit

### DIFF
--- a/backend/functions/src/middleware/auth-middleware.ts
+++ b/backend/functions/src/middleware/auth-middleware.ts
@@ -45,49 +45,25 @@ export function checkAuth(req: Request, res: Response, next: NextFunction) {
 }
 
 export function logReqBody(req: Request, res: Response, next: NextFunction) {
-  // TODO (richardgu): find a way to also log which endpoint is being hit
-  let output = `Request Endpoint: ${req.originalUrl}  `
-  if (Object.keys(req.params).length > 0) {
-    output += 'REQUEST PARAMS: {'
-    let ind = 0
-    for (const key in req.params) {
-      const val = req.params[key]
-      output += `${JSON.stringify(key).substring(
-        1,
-        JSON.stringify(key).length - 1
-      )}: ${JSON.stringify(val).substring(1, JSON.stringify(val).length - 1)}`
-      if (ind < Object.keys(req.params).length - 1) output += ', '
-      ind++
-    }
-    output += '}'
-  } else {
-    output += 'Request params is empty.  '
-  }
-  if (Object.keys(req.body).length > 0) {
-    output += 'REQUEST BODY: {'
-    let ind = 0
-    for (const key in req.body) {
-      const val = req.body[key]
-      output += `${JSON.stringify(key).substring(
-        1,
-        JSON.stringify(key).length - 1
-      )}: ${JSON.stringify(val).substring(1, JSON.stringify(val).length - 1)}`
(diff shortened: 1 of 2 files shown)


backend/functions/src/middleware/auth-middleware.ts
```diff
@@ -43,10 +44,50 @@ export function checkAuth(req: Request, res: Response, next: NextFunction) {
   }
 }
 
+export function logReqBody(req: Request, res: Response, next: NextFunction) {
+  // TODO (richardgu): find a way to also log which endpoint is being hit
```
mluo24: You can get rid of this TODO comment.

backend/functions/src/middleware/auth-middleware.ts
```diff
@@ -43,10 +44,50 @@ export function checkAuth(req: Request, res: Response, next: NextFunction) {
...
+        1,
+        JSON.stringify(key).length - 1
+      )}: ${JSON.stringify(val).substring(1, JSON.stringify(val).length - 1)}`
+      if (ind < Object.keys(req.params).length - 1) output += ', '
+      ind++
+    }
+    output += '}'
+  } else {
+    output += 'Request params is empty.  '
+  }
```
mluo24: I don't think you need these sets of if/elses. If the params are empty, I think you can just use a visual reqpresentation of something being empty, like `{}`.

mluo24 (changes requested):
Good job on implementing logging to the console! This information will definitely be more useful for keeping track of what requests were made for debugging/tracking purposes. I commend your use of middleware so you didn't have to add this code to every function.

I know you had an issue with adding new lines, and I'm pretty sure it's just because JSON doesn't work like that. I'm not sure if you really looked into much about Firebase logging, but I think there are more useful formats for logging rather than just putting everything into strings. Try experimenting with passing an entire JSON object in, as it will probably be more readable: https://firebase.google.com/docs/functions/writing-and-viewing-logs#custom-logs

Also, I recommend adding request type too, meaning whether it's `POST`, `GET`, etc.

Otherwise, your methods for approaching this are solid!

mluo24 (approved):
Thanks again for working on this! Everything looks good now, especially with the object being passed in as the argument. Definitely more informative than what we had before for logging.

One nit that I have is that params have been stringified, unlike the body, which is still a JavaScript object, but since there don't seem to be examples where there are non-empty parameters, I think this is fine.

Approving this. :)

PR #110, opened 2022-10-13
Disable match button

b3njamint:
### Summary 

This pull request disables the match button when where no groups can be formed (only 0 or 1 unmatched students). It also changes its color to visually show the button is disabled.

- [x] implemented X
- [ ] fixed Y

### Test Plan 

2 or more unmatched students: dark green
Screen Shot 2022-10-13 at 6 44 43 PM
1 unmatched student: very light black
Screen Shot 2022-10-13 at 6 44 57 PM
0 unmatched students: very light black
Screen Shot 2022-10-13 at 6 45 04 PM

### Notes 

MatchButton and how it is used seem to be a little weird in that the regular ways of changing button styling don't seem to work.

### Breaking Changes 

Nothing.


frontend/src/modules/Core/Constants/Theme.ts
```diff
@@ -19,6 +19,7 @@ export const colors = {
   lightgreen: '#CEF5D6',
   darkgreen: '#157E2C',
   darkgray: '#5C5B6A',
+  verylightblack: '##DBDBDD',
```
mluo24, line 22: You have two '#' here (though you shouldn't need this color anymore, unless you wish to use it)

frontend/src/modules/EditZing/Components/UnmatchedGrid.tsx
```diff
@@ -43,7 +43,11 @@ export const UnmatchedGrid = ({
           <StyledUnmatchedText>
             Unmatched Students ({unmatchedStudents.length})
           </StyledUnmatchedText>
-          <MatchButton label="Match" onClick={handleMatchStudents} />
+          <MatchButton
+            label="Match"
+            onClick={matchStudents}
+            disabled={unmatchedStudents.length < 2}
+          />
```
mluo24, line 50: replace MUI button here, get rid of `MatchButton` reference entirely, and it should still have the functionality.

mluo24 (changes requested):
Good work on this fix Benji! Sorry for taking so long before getting to it! After thinking about it some more, I recommend that instead of `MatchButton`, you use a regular [MUI button](https://mui.com/material-ui/react-button/) as part of cleaning up the code from the previous component library (that we are trying to move away from). That way, you can just change the `backgroundColor` and hover color (using `'&:hover'` as an object, you can find examples in the theme with `styleOverrides`) and a bit of styling if you want with `sx` as opposed to using the redundant component, but the disabling color should be done automatically in this case when you set the disabled field so you don't need to add disabled colors anymore. 

These fixes should be pretty quick and look cleaner/result in fewer redundant changes. Thanks again for working on this!

Also, while you're at it, please pull from main to resolve merge conflicts!

mluo24:
Please close this PR since you opened a new one :) @b3njamint

PR #111, opened 2022-10-13
fixed ordering of groups to numerical order

JustKong13:
### Summary 

These changes fixes the ordering of groups from alphabetical ordering to numerical ordering. The main issue previously was that when the groups were first formed, groups were formed in numerical order, however in Firebase it was stored in alphabetical order.  When it is called upon loading the page, it is called in the order that exists in Firebase, which is alphabetical rather than in numerical order. 

### Test Plan 

image
Groups now go from [1 -> 2 -> ... -> 9 -> 10 -> 11] instead of [1 -> 10 -> 11 -> ... -> 2]

image
The students in each group was also made sure to be the same 

image
Unmatching and reassigning students still keeps wanted order upon reload

### Notes 

A copy of studentGroups, [copyStudentGroups] was created and sorted, and that is what's being mapped in the frontend. This hopefully shouldn't cause any issues since studentGroups still exists


Coreball (approved):
Good solution! Nice, clean, and doesn't rely on remembering any state or manually keeping anything updated.

PR #113, opened 2022-10-14
Preserve filter and sorting choices

jewang25:
### Summary 

Preserve filter and sorting choices when navigating back and forth from dashboard page. It should last when you press both the back button on the browser and the dashboard icon. 

This pull request is the first step towards implementing feature preserving filter and sorting options

- implemented filtering and sorting the order of course cards on dashboard so that it would stay the same when navigating to and from the dashboard
- fixed code quality from last pull request related to implementing the filter options for the courses

### Test Plan 

Added various classes and changed from different filtering and sorting choices to see if the order of the course cards on the dashboard was preserved after moving from the view course page to the dashboard.

### Notes 

Used history and state to keep track of what user selected.

https://user-images.githubusercontent.com/94809605/195739368-2e1e6c78-7139-433c-9ea8-6c0b92540973.mov


frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -28,17 +28,34 @@ type FilterOption =
...
 export const Dashboard = () => {
+  const history = useHistory()
   const { user } = useAuthValue()
   const { courses } = useCourseValue()
   const { students } = useStudentValue()
-
-  const [filteredOption, setFilteredOption] = useState<FilterOption>(
-    'no-filter'
+  const state = history.location.state as {
+    sortedOrder: SortOrder
```
rgu0114, line 54: Great use of history here to store what you had for sort order and filter on the dashboard! It seems to work really well based off the videos you attached for testing. I also like what you did below by mapping elements of filterOptionDisplay instead of hard-coding the menu items – nice work there as well.

frontend/src/modules/Dashboard/Components/CourseCard.tsx
```diff
@@ -14,9 +14,17 @@ export const CourseCard = ({
   groupsFormed,
 }: CourseCardProps) => {
   const history = useHistory()
-
   const handleClickView = () => {
-    history.push(`${EDIT_ZING_PATH}/${id}`)
+    const state = history.location.state as any
```
mluo24, line 18: Get rid of the any here, I think you defined a concrete type elsewhere when you called this.

frontend/src/modules/EditZing/Components/EditZing.tsx
```diff
@@ -46,7 +46,8 @@ const LscIcon = (props: SvgIconProps) => {
 export const EditZing = () => {
   const { courseId } = useParams<{ courseId: string }>()
   const [showError, setShowError] = useState(false)
-
+  const history = useHistory()
+  const state = history.location.state as any
```
mluo24, line 37: same here

frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -28,17 +28,34 @@ type FilterOption =
...
+  ['newly-matchable', 'Newly Matchable'],
+  ['matchable', 'Matchable'],
+  ['no-check-in-email', 'No Check In Email'],
+  ['no-no-match-email', 'No No Match Email'],
+]
+const sortOrderDisplay = [
+  ['newest-requests-first', 'Newest Requests First'],
+  ['classes-a-z', 'Classes A-Z'],
+  ['classes-z-a', 'Classes Z-A'],
+]
```
mluo24, line 46: I think you can use a different data structure for this, maybe a Map instead of an associative list. I think for this purpose, it doesn't make any difference since you're iterating through everything anyway, but in the future, the mapping could be useful. This is an easy fix, as I think Maps can just take in associative lists for initialization too.

frontend/src/modules/Dashboard/Components/CourseCard.tsx
```diff
@@ -14,9 +14,17 @@ export const CourseCard = ({
...
+    history.push({
+      pathname: `${EDIT_ZING_PATH}/${id}`,
+      state: {
+        sortedOrder: state?.sortedOrder
+          ? state.sortedOrder
+          : 'newest-requests-first',
+        filterOption: state?.filterOption ? state.filterOption : 'no-filter',
+      },
```
mluo24: same here for this about the constants, probably easier to change across files too.

frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -28,17 +28,34 @@ type FilterOption =
...
+    filterOption: FilterOption
+  }
+  const [sortedOrder, setSortedOrder] = useState<SortOrder>(() =>
+    state?.sortedOrder ? state.sortedOrder : 'newest-requests-first'
   )
-  const [sortedOrder, setSortedOrder] = useState<SortOrder>(
-    'newest-requests-first'
+  const [filteredOption, setFilteredOption] = useState<FilterOption>(() =>
+    state?.filterOption ? state.filterOption : 'no-filter'
   )
```
mluo24: Super specific nit, but for better readability/easier changes, you can make constant variables for "default sorting order" and "default no filter" in place of the "newest-requests-firsts" and "no-filter"

mluo24 (approved):
Thank you Jessica for implementing this, and nice work! Your solution is complete and elegant, and I also appreciate your use of using history state to handle this! This loses the functionality of having a bookmarkable URL, but this accomplishes the same thing but more, as well as having a way to preserve the state when just clicking the icon button to go back.

By the way, in the future, I recommend separating fixes in PRs, since I know you fixed some things from your previous PR in this one, but I recommend just opening another, separate PR. That way, you get your PR count up and it's easier to review for everyone.

Also, remember to pull main before merging to resolve any merge conflicts!

PR #114, opened 2022-10-20
line breaks show on same line when opening csv file

JustKong13:
### Summary 

This pull request fixes the bug where there are actual line breaks when viewing the file rather than `\n`. 

### Test Plan 
Previously, when there were multiple line breaks, it would take up multiple lines in the csv file
Untitled

Now, they are separated by `\n`
Untitled

### Notes 
Only thing that might be of concern is if you were viewing the graphical version of the csv, you would see `\n` rather than the line breaks
image

As of right now, I don't think there's anyway to show the global modifier on the csv file but not in excel since that would depend on your OS

### DIFF
--- a/frontend/src/modules/Dashboard/Components/Dashboard.tsx
+++ b/frontend/src/modules/Dashboard/Components/Dashboard.tsx
@@ -90,7 +90,7 @@ export const Dashboard = () => {
                   : undefined,
               ...localeMap(group?.templateTimestamps),
               ...localeMap(membership.templateTimestamps),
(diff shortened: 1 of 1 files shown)


b3njamint (commented):
Good job on your fix! I appreciate how you simplified the problem and solved it with a very *concise* solution. Not sure if the "\n" in the csv files would cause any problems on the user's end?

mluo24 (approved):
I think your description may be out of date since you just replace it with double spaces now, but looks good to me! Using replace and regex along with trimming in the case of a blank note or random extra spaces is nice and concise solution. Thanks for working on this, and the CSVs should look much cleaner now.

PR #115, opened 2022-10-23
Background Refresh – regular prompts for users 

rgu0114:
### Summary 
Updated reload message look:
https://user-images.githubusercontent.com/100158738/200981729-87b67161-edc4-43cd-84bb-bcb383df3d94.mov

Reload prompt appears when the number of groups has changed: 
https://user-images.githubusercontent.com/100158738/201234848-c4b9c9ff-994a-404f-956e-54465fed62ff.mov

Reload prompt appears when the group composition changes: 
https://user-images.githubusercontent.com/100158738/201234903-3b572908-0c49-4c94-81c3-3e2bd700ab1f.mov

_____________________________________________________________________________________________________

This pull request is the first step towards implementing messages prompting users to reload the page after a certain period of time (see [Notion task](https://www.notion.so/Background-refresh-for-new-updates-in-centralized-state-2b1e379a681742ef8677b5901ebaa02d) for more details). These changes are necessary as only hard refreshes will show changes to the new centralized state. 

**Implementation Details**

- I used setInterval inside a useEffect to start a timer for how long the user has been on the site without refreshing, starting on the first load. When the interval is reached, we first check to see if our data has changed by refetching courses and students and comparing them with the currently existing state. If there's a noticeable difference, we fire off a function that renders a popup prompting users to reload the page. 
- Placed this logic near the root of the app to (hopefully) have this apply to all pages, instead of page by page. This appears to be working so far on the pages I've tested. Hopefully it stays that way. 
- Created messaging using React MUI's snackbar alert

### Test Plan 

(Note that the current time intervals are just for temporary, will eventually raise to ~15 minutes I think) 

10/27: Popup only appears if there is a data change (change in students after submitting the survey) 

https://user-images.githubusercontent.com/100158738/198424648-5f6a9b39-85aa-453a-b908-29d172c8d162.mov

Popup appearing after ~10 sec on the dashboard page and disappearing after several seconds: 

https://user-images.githubusercontent.com/100158738/197626380-c3324442-2d3f-4f69-a643-8da77a59b3ca.mov

On a class-specific page:

Screen Shot 2022-10-24 at 4 46 55 PM

### Notes 

- Message color is a bit ugly.


JustKong13:
Nice work! The little message pop up at the bottom seems to be a lot shorter than 10 seconds (I think I counted ~3-5 seconds), but it might be an issue that if you set it  for 10 seconds and it's less than 10 seconds, maybe the timing is off? Not really sure. 

Also for me, the popup is not disappearing a few seconds of it being up, which you mentioned. It's not a breaking change but it's just something to note since it's not behaving 100% the way you want it to be. 

Lastly, pressing the reload button doesn't reload the page for me.

frontend/src/App.tsx
```diff
@@ -40,21 +41,62 @@ import { Emailing } from 'Emailing'
...
+import React from 'react'
 
 const App = () => {
   const [currentUser, setCurrentUser] = useState<User | null>(null)
   const [authState, setAuthState] = useState<AuthState>('loading')
   const axiosAuthInterceptor = useRef<number | null>(null)
   const [networkError, setNetworkError] = useState<string | null>(null)
 
+  const [needsRefresh, setNeedsRefresh] = useState(false)
+  const checkRefreshDuration = 10000 //600000 (10 min)
```
mluo24: Remember to change this line!

frontend/src/App.tsx
```diff
@@ -40,21 +41,62 @@ import { Emailing } from 'Emailing'
...
+    <Alert
+      onClose={handleClose}
+      severity="info"
+      sx={{ width: '100%' }}
+      variant="filled"
+      action={
+        <Button
+          variant="contained"
+          size="small"
+          // onClick={document.location.reload()}
```
mluo24: Remember to put this back in, or find a way to get the page to hard refresh.

frontend/src/App.tsx
```diff
@@ -40,21 +41,62 @@ import { Emailing } from 'Emailing'
...
+        >
+          Reload
+        </Button>
+      }
+    >
+      You haven't reloaded in a while and there may be updated information
```
mluo24: I think we can make this change this to say something like "The information in your app is out of date. Please reload to see the latest updates"

frontend/src/App.tsx
```diff
@@ -137,6 +179,37 @@ const App = () => {
...
 
+  useEffect(() => {
+    if (hasLoadedCourses && hasLoadedStudents) {
+      console.log('setting interval')
+      const interval = setInterval(async () => {
+        console.log(
+          `${
+            checkRefreshDuration / 1000
+          } seconds have passed — now checking if course/student data have changed.`
+        )
```
mluo24: Remember to get rid of console.logs for when you merge into main.

frontend/src/App.tsx
```diff
@@ -40,21 +41,62 @@ import { Emailing } from 'Emailing'
...
+  const handleClose = (
+    event?: React.SyntheticEvent | Event,
+    reason?: string
+  ) => {
+    if (reason === 'clickaway') {
+      return
+    }
+
+    setNeedsRefresh(false)
+  }
```
mluo24: Is this to close the pop up when you just click on the screen? if so, I don't think you need this.

frontend/src/App.tsx
```diff
@@ -40,21 +41,62 @@ import { Emailing } from 'Emailing'
 import { TemplateEditor } from 'TemplateEditor'
 import './App.css'
 import theme from '@core/Constants/Theme'
-import { User, onAuthStateChanged } from 'firebase/auth'
+import { User, onAuthStateChanged, reload } from 'firebase/auth'
```
mluo24: probably get rid of this because of linting errors.

frontend/src/App.tsx
```diff
@@ -40,21 +41,62 @@ import { Emailing } from 'Emailing'
...
+        >
+          Reload
+        </Button>
+      }
+    >
+      You haven't reloaded in a while and there may be updated information
+    </Alert>
+  )
+  const [refetchedCourses, setRefetchedCourses] = useState<Course[]>([])
+  const [refetchedStudents, setRefetchedStudents] = useState<Student[]>([])
```
mluo24: I don't think this was used (or necessary), so you can remove it.

mluo24 (changes requested):
Good job with implementing this feature! I think this is important for the new centralized state structure of the app and will improve user experience overall. 

image
I also was finally able to find the component for a notification, which can also be found in the [Figma](https://www.figma.com/file/5sae0s8rk6r9iVwpn74RY4/Zing-Components?node-id=312%3A74). I think there are some examples already in the codebase, including for when you add a note, for example. 

I also think that we need to account for groups being made as part of checks for reloading in addition to classes and students, meaning if the groups are equal or not (because groups are dynamic and can change). I don't think there are any other considerations (updates to emails are very not frequent from my understanding).

Once those little changes are made I think everything will be good!

frontend/src/App.tsx
rgu0114: yee, oke

frontend/src/App.tsx
```diff
@@ -48,13 +49,36 @@ import { getDownloadURL, ref } from 'firebase/storage'
...
+    reason?: string
+  ) => {
+    if (reason === 'clickaway') {
+      return
+    }
+  }
+
+  function refreshPage() {
+    window.location.reload()
+  }
```
mluo24: I think you can just get rid of this and put `window.location.reload()` directly in the button action.

frontend/src/App.tsx
```diff
@@ -137,6 +161,52 @@ const App = () => {
     )
   }
 
+  useEffect(() => {
+    if (hasLoadedCourses && hasLoadedStudents) {
```
mluo24, line 157: I think it's worth having a check for the emails being loaded in as well, since if it wasn't there, then the snackbar could pop up during the loading screen if data happened to have changed there. Visually, I think that may be weird.

frontend/src/App.tsx
```diff
@@ -137,6 +161,52 @@ const App = () => {
...
+          await axios.get(`${API_ROOT}${COURSE_API}`)
+        ).data.map(responseCourseToCourse)
+
+        const newStudents = (
+          await axios.get(`${API_ROOT}${STUDENT_API}`)
+        ).data.map(responseStudentToStudent)
+
+        newClasses.forEach((course: Course, index: number) => {
+          let newGroups = course.groups
+          let oldGroups = courses[index].groups
```
mluo24: If there are no groups made, the `groups` property can be undefined, therefore throwing an error. You should have a condition to check if it exists, otherwise just return an empty array (because then there are no groups) and the rest of the code should still work.

mluo24 (commented):
Hi Richard, thanks for addressing all the changes! It looks like everything works except for when there are no groups made for a certain class, as then the properties will be undefined. To reproduce, load the app, then run `npm run populate` to populate the classes, and then wait 10 minutes (or however long the interval is).

image

This should be a simple fix, and the last of them (as long as we've exhausted the important cases, which tbh I think we have)! I will also be adding Changyuan to review once you're done so he can evaluate how your code works.

Also, please fix linting errors.

frontend/src/App.tsx
rgu0114: o true

frontend/src/App.tsx
rgu0114: If a class doesn't have any groups, wouldn't the `groups` property be initialized to the empty array already?

frontend/src/App.tsx
rgu0114: remarking as unresolved until tmr bc the change makes it unresponsive to cases where the snackbar should be showing up for some reason

mluo24: Oh interesting, not sure why that would be the case.

frontend/src/App.tsx
mluo24: I thought so too, but if that were the case, this error wouldn't have been thrown. You can print out the whole object to see if what you're saying is true.

rgu0114: Oh I think the error was bc of a mismatch between the old # of classes and the new, so I couldn't do courses[index] in all cases. I switched some stuff around to account for that and it seems to have gone away. All the classes with no groups I've tried printing out have had [] and not null.

frontend/src/App.tsx
rgu0114: Maybe templates just taking a long time to load? I've been having some issues with templates loading recently

frontend/src/App.tsx
mluo24: ok makes sense, yeah I guess even TypeScript is still weird in handling index out of bounds errors for arrays. Very nice!

frontend/src/App.tsx
mluo24: The condition should be true once the app has completely loaded, so maybe it's something else. But tbh I think this is trivial bc there's a 10 minute timer anyway, and there's no way that email templates will take 10 minutes to load. I would say disregard unless it causes issues.

frontend/src/App.tsx
```diff
@@ -48,13 +49,38 @@ import { getDownloadURL, ref } from 'firebase/storage'
...
 
 const App = () => {
   const [currentUser, setCurrentUser] = useState<User | null>(null)
   const [authState, setAuthState] = useState<AuthState>('loading')
   const axiosAuthInterceptor = useRef<number | null>(null)
   const [networkError, setNetworkError] = useState<string | null>(null)
 
+  const [needsRefresh, setNeedsRefresh] = useState(false)
+  const checkRefreshDuration = 1000
+  // 600000 // (10 min)
```
Coreball: Remember to set this to the production value before merging

frontend/src/App.tsx
```diff
@@ -48,13 +49,38 @@ import { getDownloadURL, ref } from 'firebase/storage'
...
+
+  const handleClose = (
+    event?: React.SyntheticEvent | Event,
+    reason?: string
+  ) => {
+    if (reason === 'clickaway') {
+      return
+    }
+  }
```
Coreball: I don't think you need this function anymore, to do nothing simply don't pass Snackbar the `onClose`prop

frontend/src/App.tsx
```diff
@@ -137,6 +163,54 @@ const App = () => {
...
+  useEffect(() => {
+    if (hasLoadedCourses && hasLoadedStudents) {
+      const interval = setInterval(async () => {
+        const newClasses = (
+          await axios.get(`${API_ROOT}${COURSE_API}`)
+        ).data.map(responseCourseToCourse)
+
+        const newStudents = (
+          await axios.get(`${API_ROOT}${STUDENT_API}`)
+        ).data.map(responseStudentToStudent)
```
Coreball: Some error handling would be nice, e.g. if they lose internet connectivity and we're unable to check if there were updates. I think you can use `setNetworkError()` which will make the snackbar for that appear automatically.

frontend/src/App.tsx
```diff
@@ -137,6 +163,54 @@ const App = () => {
...
+                  setNeedsRefresh(true)
+                }
+              })
+            })
+          } else {
+            setNeedsRefresh(true)
+          }
+        })
+      }, checkRefreshDuration)
+      return () => clearInterval(interval)
```
Coreball, line 190: Nice use of the cleanup function of useEffect

frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
```
Coreball: Remove this change because it's causing the lint error

frontend/src/App.tsx
Coreball: For terminology purposes it would also be good to call it newCourses instead of newClasses

Coreball (commented):
Good work Richard! This will improve the user experience by giving them a way of knowing when there have been changes in the storage backing the app. I see that you have already addressed a lot of Miranda's feedback, which is great.

During my testing, I encountered a situation where the app doesn't display the refresh snackbar even though new surveys have been submitted. This happens when a student that has already submitted surveys submits the survey for an existing course that they aren't a part of yet. The if statement on line 177 doesn't catch this because the lengths of the new courses and the new students are unchanged, and the group membership checking doesn't catch this because the members of all groups are unchanged.

Another way that you could consider implementing the comparison between the existing data and the new data is by taking a student-oriented approach, instead of the course-oriented approach. Courses have a list of unmatched students and a group list where each group contains members, which is a bit more complicated to reason about than students, where each student has group memberships containing the course ID and their group number. You may actually be able to get away with only calling the students endpoint in the student-oriented approach, because if any new courses appear it also means that a new group membership object has been added somewhere. If you just compare the group number to the previous group number it should cover when the student is unmatched as well since it's defined as -1.

mluo24 (approved):
lgmt

PR #116, opened 2022-10-27
Fix export noncurrent semesters

JustKong13:
### Summary 

This pull request makes it so that when you export student data, or course data, as a csv, you will only export the data for this current semester. This is done by filtering out the data to be equal to the current semester

### Test Plan 

Previously, this is what exporting the courses would look like
image

Now, our spring semester for CS2110 is gone
image

The same thing happens for students. Now, only the current semester is displayed
image

When creating dummy data, I momentarily changed the source code to randomly select between "FA22" and "SP22" rather than hardcoding in "FA22"

### DIFF
--- a/frontend/src/modules/Dashboard/Components/Dashboard.tsx
+++ b/frontend/src/modules/Dashboard/Components/Dashboard.tsx
@@ -293,7 +293,10 @@ export const Dashboard = () => {
             horizontal: 'right',
           }}
         >
-          <CSVLink data={csvCourses} filename={`export-courses-${Date.now()}`}>
+          <CSVLink
+            data={csvCourses.filter((e) => e.semester === 'FA22')}
+            filename={`export-courses-${Date.now()}`}
+          >
             <MenuItem>Export CSV (Courses)</MenuItem>
           </CSVLink>
           <CSVLink
--- a/frontend/src/modules/Dashboard/Components/Dashboard.tsx
+++ b/frontend/src/modules/Dashboard/Components/Dashboard.tsx
@@ -294,13 +294,13 @@ export const Dashboard = () => {
           }}
         >
           <CSVLink
-            data={csvCourses.filter((e) => e.semester === 'FA22')}
+            data={csvCourses.filter((e) => e.semester === selectedRoster)}
             filename={`export-courses-${Date.now()}`}
           >
             <MenuItem>Export CSV (Courses)</MenuItem>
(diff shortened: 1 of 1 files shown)


jewang25 (commented):
Your change looks good! I like how simple you kept your changes so that you didn't change much of the existing code. I also like how you mentioned that you tested with both FA22 and SP22 by changing the source code to randomly select between the two because it is important to test all possibilities.

frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -293,11 +293,14 @@ export const Dashboard = () => {
             horizontal: 'right',
           }}
         >
-          <CSVLink data={csvCourses} filename={`export-courses-${Date.now()}`}>
+          <CSVLink
```
rgu0114, line 296: I really like how concisely you were able to make this fix + efficiently used available functions to minimize the disruption to existing code. This works very well, based off what you demoed during work session yesterday. Great work!

frontend/src/modules/Dashboard/Components/Dashboard.tsx
```diff
@@ -293,11 +293,14 @@ export const Dashboard = () => {
...
+            data={csvCourses.filter((e) => e.semester === 'FA22')}
+            filename={`export-courses-${Date.now()}`}
+          >
             <MenuItem>Export CSV (Courses)</MenuItem>
           </CSVLink>
           <CSVLink
-            data={csvStudents}
+            data={csvStudents.filter((e) => e.semester === 'FA22')}
```
Coreball: I think you can use `selectedRoster` instead of hard coding the string `FA22`. Surprised that no one else has mentioned this

Coreball (commented):
Good work, I like how you were able to test this by creating survey submissions for multiple semesters. I also like the approach of filtering at the end after the data is mapped to CSV rows instead of trying to filter during the row generation. This keeps it simple and easy to understand, since it's a lot like filtering in a spreadsheet.

I would suggest using the variable for selected semester instead of hard coding it. This means that when the Spring 2022 semester (or any other semester) is selected, the exported CSV will correspond to what's currently being shown on the dashboard instead of being for the Fall 2022 semester only.

mluo24 (approved):
All of the concerns have been addressed, and it looks complete now. Nice work, lgtm!

PR #118, opened 2022-11-15
Disable match button 2

b3njamint:
### Summary 

The SQL (sequel) to #110 

This pull request disables the match button when where no groups can be formed (only 0 or 1 unmatched students). It also changes its color to visually show the button is disabled.

- [x] implemented X
- [ ] fixed Y

### Test Plan 

Initial state: match button is green!
Screen Shot 2022-11-14 at 7 57 52 PM

After matching and there are 0 people left to be matched: match button is very light black!
Screen Shot 2022-11-14 at 7 58 07 PM

When there is 0 or 1 people, then the match button is still very light black because no groups can be formed.
Screen Shot 2022-11-14 at 7 58 17 PM

### Notes 

To follow @mluo24's suggestions from the previous PR, I used a regular MUI button instead of a MatchButton (on it's way to depreciation).

### Breaking Changes

### DIFF
--- a/frontend/src/modules/Core/Constants/Theme.ts
+++ b/frontend/src/modules/Core/Constants/Theme.ts
@@ -19,7 +19,6 @@ export const colors = {
   lightgreen: '#CEF5D6',
   darkgreen: '#157E2C',
   darkgray: '#5C5B6A',
-  verylightblack: '#DBDBDD',
 }
 
 // module augmentation, needed for TypeScript
--- a/frontend/src/modules/Core/Constants/Theme.ts
+++ b/frontend/src/modules/Core/Constants/Theme.ts
@@ -19,6 +19,7 @@ export const colors = {
   lightgreen: '#CEF5D6',
   darkgreen: '#157E2C',
   darkgray: '#5C5B6A',
+  verylightblack: '#DBDBDD',
 }
 
 // module augmentation, needed for TypeScript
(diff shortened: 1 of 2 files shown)


frontend/src/modules/Core/Constants/Theme.ts
```diff
@@ -19,6 +19,7 @@ export const colors = {
   lightgreen: '#CEF5D6',
   darkgreen: '#157E2C',
   darkgray: '#5C5B6A',
+  verylightblack: '#DBDBDD',
```
mluo24: This color isn't needed.

frontend/src/modules/EditZing/Components/UnmatchedGrid.tsx
```diff
@@ -43,7 +43,22 @@ export const UnmatchedGrid = ({
...
-          <MatchButton label="Match" onClick={handleMatchStudents} />
+          <Button
+            variant="contained"
+            onClick={handleMatchStudents}
+            disabled={unmatchedStudents.length < 2}
+            sx={{
+              backgroundColor: colors.darkgreen,
+              disabled: {
+                backgroundColor: colors.verylightblack,
+              },
```
mluo24: This also isn't needed. It's default grey if it's disabled.

mluo24 (approved):
Thanks for the quick and simple fix Benji! Also thanks for incorporating all of my feedback from the last PR. I'm approving this as it looks good to me, but I added a few comments that you can address to the code (it's just removing some unnecessary lines). Make sure to close the previous PR too. Good work :)

PR #119, opened 2022-11-15
Non-"unmatchable" filters don't contain classes with unmatchable students 

rgu0114:
### Summary 

Currently, some of the filters (like matchable or send email) contain classes with unmatchable students (0 groups, 1 student). This PR resolves this issue by adding an extra layer of filtering for this specific case. Now, only "unmatchable" and "no filter" contain classes with unmatchable students.

### Test Plan 

CS 3110 (unmatchable since it only has one student) only appears when no filter or "unmatchable" is selected, and doesn't appear in any of the other filters.

https://user-images.githubusercontent.com/100158738/202047032-4833e6fc-e3ad-42c9-99f2-8d34b3d42b54.mov


mluo24:
Hi Richard, thanks for taking the time to work on this task. I'm sorry that I didn't get to this earlier! I think there may have been a misunderstanding of what your task was ("Un-matchable vs. can add to already existing group") here, and it was to add a new filter in the dropdown for "Can add to already existing group." You didn't have to modify any of the existing filters. Because of large differences between this PR and the task description, and also since this branch is kind of out of date with main, I'm going to close this PR so you can make a new one with your updated changes.

rgu0114:
> Hi Richard, thanks for taking the time to work on this task. I'm sorry that I didn't get to this earlier! I think there may have been a misunderstanding of what your task was ("Un-matchable vs. can add to already existing group") here, and it was to add a new filter in the dropdown for "Can add to already existing group." You didn't have to modify any of the existing filters. Because of large differences between this PR and the task description, and also since this branch is kind of out of date with main, I'm going to close this PR so you can make a new one with your updated changes.

Oops mb, I should've double checked the task description with you earlier. I'll get started on the changes later today.

PR #120, opened 2022-11-17
Delete empty groups

JustKong13:
### Summary 

This change gives users the ability to delete empty groups (although the group will still exist, it just will not be shown to user)

- [x] Change database to take into account a new field `hidden`
- [x] Frontend only displays groups with `hidden` field of `false`
- [x] Backend implementation to handle changes with `hidden`
- [x] Styling with the garbage can and checkbox depending on the population of group

https://user-images.githubusercontent.com/64031655/202386745-32e5df48-bdc6-4ed4-90e8-d66b05ab6478.mp4

### Test Plan 

Principles of test driven development (TDD) was used throughout the implementation of this feature because there was a lot of different parts of the code base that needed editing. This is roughly the sequence that each feature was developed as well. 

1. Testing of the `hidden` field (from changing type definitions) was done by simply looking at the database
2. We wanted only groups with `{ hidden: false }` to show up. This was tested by manually changing the field in the database, and seeing whether that group appears in the course page
3. After the backend helper function `hideEmptyGroup` and it's associated endpoint `matching/hide-group` was written, this was tested by making API calls using Insomnia, and observing whether a group's `hidden` state switched from `false` to `true`.
4. Integration with the backend to the frontend was mainly done through visual testing (see video above) 

### Notes 

- We probably don't want the page to fully reload every time a group had been deleted
- Definitely should add a confirmation button to confirm deleting a group
- Axios, `API_ROOT`, and `MATCHING_API` are all called in `GroupCard.tsx`. Scanning the other parts of the database, the majority of API calls are written in `App.tsx`. However I suspect that was because we wanted to centralize the movement of students for the entire codebase. In addition, I don't expect the added backend functions and endpoints to be called anywhere else but `GroupCard.tsx`, which is why I choose to put it there instead. 
- I also had to reclone this repo, so don't mind the 14k insertions and deletions

### Breaking Changes 

The majority of breaking changes are where types are defined. There were quite a number of types that needed to be changed in order for this feature to work properly.
- In `./backend/functions/src/matching/functions.ts`, `groupCreationUpdate`, `newGroup`, `newGroups`, `newGroupsFirestore` added an additional field `hidden: false` to initialize a new group as not being hidden
- In `./backend/functions/src/types/index.ts`, types needed to be changed. `Group`, and `FirestoreGroup` were updated to have `hidden: boolean`
- In `./backend/functions/src/course/functions/ts`, `getAllCourses` was changed so that `groups` in `courseData` will also allow us to access `hidden`


Coreball (changes requested):
I am not reviewing this yet because the diff is way too large and contains many unrelated changes. You should close this PR and redo the things related to the topic of this pull request on a clean branch based off of the most recent version of main, since to my understanding this doesn't depend on any of the other commits authored by other people in this PR? If we merged this pull request, the commit history would be polluted and it'd be best to avoid that.

PR #121, opened 2022-11-17
implemented option to delete groups

JustKong13:
### Summary 

This change gives users the ability to delete empty groups (although the group will still exist, it just will not be shown to user)

- [x] Change database to take into account a new field `hidden`
- [x] Frontend only displays groups with `hidden` field of `false`
- [x] Backend implementation to handle changes with `hidden`
- [x] Styling with the garbage can and checkbox depending on the population of group

https://user-images.githubusercontent.com/64031655/202386745-32e5df48-bdc6-4ed4-90e8-d66b05ab6478.mp4

### Test Plan 

Principles of test driven development (TDD) was used throughout the implementation of this feature because there was a lot of different parts of the code base that needed editing. This is roughly the sequence that each feature was developed as well. 

1. Testing of the `hidden` field (from changing type definitions) was done by simply looking at the database
2. We wanted only groups with `{ hidden: false }` to show up. This was tested by manually changing the field in the database, and seeing whether that group appears in the course page
3. After the backend helper function `hideEmptyGroup` and it's associated endpoint `matching/hide-group` was written, this was tested by making API calls using Insomnia, and observing whether a group's `hidden` state switched from `false` to `true`.
4. Integration with the backend to the frontend was mainly done through visual testing (see video above) 

### Notes 

- We probably don't want the page to fully reload every time a group had been deleted
- Definitely should add a confirmation button to confirm deleting a group
- Axios, `API_ROOT`, and `MATCHING_API` are all called in `GroupCard.tsx`. Scanning the other parts of the database, the majority of API calls are written in `App.tsx`. However I suspect that was because we wanted to centralize the movement of students for the entire codebase. In addition, I don't expect the added backend functions and endpoints to be called anywhere else but `GroupCard.tsx`, which is why I choose to put it there instead. 
- There's also a deploy to Firebase test that is failing, probably because we've changed the contents of the firebase.

### Breaking Changes 

The majority of breaking changes are where types are defined. There were quite a number of types that needed to be changed in order for this feature to work properly.
- In `./backend/functions/src/matching/functions.ts`, `groupCreationUpdate`, `newGroup`, `newGroups`, `newGroupsFirestore` added an additional field `hidden: false` to initialize a new group as not being hidden
- In `./backend/functions/src/types/index.ts`, types needed to be changed. `Group`, and `FirestoreGroup` were updated to have `hidden: boolean`
- In `./backend/functions/src/course/functions/ts`, `getAllCourses` was changed so that `groups` in `courseData` will also allow us to access `hidden`

### DIFF
--- a/frontend/src/modules/EditZing/Components/GroupCard.tsx
+++ b/frontend/src/modules/EditZing/Components/GroupCard.tsx
@@ -6,7 +6,7 @@ import { STUDENT_TYPE, DnDStudentTransferType } from 'EditZing/Types/Student'
 import { StyledGroupText } from 'EditZing/Styles/StudentAndGroup.style'
 import { Box, Tooltip, Checkbox, IconButton } from '@mui/material'
 import CircleIcon from '@mui/icons-material/Circle'
-import { Delete, DeleteOutline } from '@mui/icons-material'
+import { Delete } from '@mui/icons-material'
 import axios from 'axios'
 import { API_ROOT, MATCHING_API } from '@core'
 
--- a/frontend/src/modules/EditZing/Components/GroupCard.tsx
+++ b/frontend/src/modules/EditZing/Components/GroupCard.tsx
@@ -24,6 +24,7 @@ const GroupCard = ({
   handleChecked,
   handleAddStudent,
   updateNotes,
+  removeGroups,
 }: GroupGridProps) => {
   /**
    * Helper to format the timestamp data in a way that is helpful for displaying in tooltips
@@ -68,12 +69,12 @@ const GroupCard = ({
     setIsHovering(false)
   }
 
-  const removeGroup = (groupId: string, groupNumber: number) => {
+  const removeGroup = (courseId: string, groupNumber: number) => {
     axios.post(`${API_ROOT}${MATCHING_API}/hide-group`, {
       courseId: courseId,
       groupNumber: groupNumber,
     })
-    window.location.reload() // this can be improved so page doesn't reloadA
+    removeGroups(courseId, groupNumber)
   }
 
   return (
--- a/frontend/src/modules/EditZing/Components/GroupCard.tsx
+++ b/frontend/src/modules/EditZing/Components/GroupCard.tsx
@@ -4,8 +4,11 @@ import { GroupGridProps } from 'EditZing/Types/ComponentProps'
 import { useDrop } from 'react-dnd'
 import { STUDENT_TYPE, DnDStudentTransferType } from 'EditZing/Types/Student'
 import { StyledGroupText } from 'EditZing/Styles/StudentAndGroup.style'
-import { Box, Tooltip, Checkbox } from '@mui/material'
+import { Box, Tooltip, Checkbox, IconButton } from '@mui/material'
 import CircleIcon from '@mui/icons-material/Circle'
+import { Delete, DeleteOutline } from '@mui/icons-material'
+import axios from 'axios'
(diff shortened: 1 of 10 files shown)


rgu0114 (changes requested):
Great work Justin! This is a very useful feature as I have come across the issue a few times of having an empty group that's sorta just sitting there. Your implementation for the most part looks well handled. One change that I would suggest, though, is that you give a warning to users after they click the trash can (which is very pretty btw) to confirm their actions as deleting a group is a pretty substantial change.

b3njamint:
Good job Justin! I like your detailed testing plan. Michael Clarkson would be proud :) I don't see any outstanding problems with your current code, but a few things to consider:

1. Do we want to outright delete the group, or just hide it?
2. If we choose just to hide it, would it be difficult to implement an undo feature? Or maybe, some way to bring back an accidentally deleted group. Some suggestions, but I think what you/Richard said about a confirmation would be fine too.
4. Finally, what do we want to do with the group numbering? As in, when we delete a group, should we start numbering from there again, or skip over it?

CollinWoo:
Great stuff Justin! The new trash can icon looks very nice and I like how thorough you were with testing. For updating the edit page without reloading, I would suggest updating the local state accessed from the CourseContext in addition to making an API call to modify the firebase database (similar to how the moveStudent and matchStudent functions work in App.tsx in #102). This will allow the locally stored data that is rendered on the page to be updated without refreshing.

frontend/src/modules/EditZing/Components/GroupCard.tsx
```diff
@@ -65,6 +68,14 @@ const GroupCard = ({
...
   }
 
+  const removeGroup = (groupId: string, groupNumber: number) => {
+    axios.post(`${API_ROOT}${MATCHING_API}/hide-group`, {
+      courseId: courseId,
+      groupNumber: groupNumber,
+    })
+    window.location.reload() // this can be improved so page doesn't reloadA
+  }
+
```
mluo24: Yeah, please change this part. Every other operation that `POST`s to the Firestore database doesn't require a reload. Take a look at examples App.tsx for functions that show how to do this, as well as some other components that may update the database. You will probably have to work with React Context on this one.

mluo24 (changes requested):
Thanks for working on this Justin! I like your use of the soft delete field, hidden. Having this will ensure that data won't be lost. Benji poses some good questions, but this has been determined by LSC to not have the numbers change for the group when a group in the middle gets deleted. I think your current implementation is good and makes sense. 

I agree with Richard in terms of user safety for deleting. Since you have the field instead of actually deleting, another alternative which may be just as simple is just an undo button. Maybe in the future, we can consider a "restore" option.

One glaring issue you do need to fix though is the reloading every time you delete a group. You want to make the request to the backend but also modify the state within the frontend.

There is also sometimes a bug where if you delete the group, it's still there, and you need to refresh/delete an additional time. I can't provide reproducibility steps because it seems like it's random, so just try it on a few classes and it'll probably show up. I think if you fix the reloading, it should be fine.

mluo24 (approved):
Everything looks good to me now with your fixes. Thanks again for working on this! If we want to restore groups now, having that "hidden" field will make everything a lot easier. If you don't have anything else for your dev portfolio, having a follow up with an "undo" popup can work. 

You just need to fix some linting errors (see below), merge in main to fix conflicts, and you should be good to go.

image

rgu0114 (approved):

JustKong13:
The linting issue as well as the merge conflicts have been resolved

PR #122, opened 2022-11-18
Use Google Account for Student Sign In

CollinWoo:
### Summary 

This pull request allows students to sign in using their cornell.edu account with Google Sign-in while filling out the student survey. 

- [x] Implemented design for new login verification page
- [x] Implemented google sign in for students
- [x] Implemented checking that student email ends in cornell.edu

### Test Plan 

Testing was performed by filling out surveys using the new system and verifying that they were submitted correctly (new students show up on the dashboards and can be emailed to). Additional testing was also performed using admin accounts to make sure that email sending and dashboard access still work as intended despite the new sign in method.

Screenshot 2022-11-17 at 6 58 58 PM
Screenshot 2022-11-17 at 6 59 23 PM

### Notes 

- Signing in as an admin using google authentication still gives all admin privileges, but you cannot send emails properly since the email authentication requires Microsoft
- Because of a change in the user account linking setting, getting the logged in user's email now requires a different method.  Currently, this setting change is performed by changing every instance of user?.email to the new format. This is probably not good coding style and not good for future extensibility. Suggestions on how to globally change the email access without manually changing every instance would be appreciated. I also might have missed some instances so not everything might work correctly
- The new log in verification screen is not perfectly responsive for wider/smaller screens
- There might be a bug with the authentication because it seems that sometimes people logged in as admin cannot access the dashboard, but I am not sure how to recreate the error

### Breaking Changes 

- The setting for user account linking in Firebase Authentication has been changed from "link accounts that use the same email" to "create multiple accounts for each identity provider"


jewang25 (commented):
The updated login page looks really good! I like how your testing plan is really thorough and how the new feature makes sure the student's email is a cornell email. When I was testing the first time, it wouldn't let me go to the dashboard from the survey page for some reason. I'm not sure if it's just me or a random situation, but if you have time you could look into that. However, it's minor so good job with implementing this new feature!

rgu0114 (commented):
This looks great Colin! I like how visually pleasing the new sign in page is, as well as how you check that the emails are valid Cornell emails using the existing infrastructure.

PR #124, opened 2022-11-18
Factored Out Account Menu

jewang25:
### Summary 

- made a separate component for logout button/account menu so that the metrics page could also use it without duplicating code

### Test Plan 

- testing that each button on the menu still worked (switching semesters, exporting csvs, and logging out)
- checked that the changed data from clicking a button produced the expected functionality

### Notes 

The video attached below shows how the account menu button still works like it used to.

https://user-images.githubusercontent.com/94809605/202793371-301e62b5-8296-4e9c-b44b-4b6f67b2a8d3.mov


mluo24 (approved):
Tested it and everything seems to work, so looks good to me! Thanks for cleanly factoring it out :)

PR #125, opened 2022-11-18
Metrics page

jewang25:
### Summary 

This pull request is the first step towards implementing a stats page to display the stats for LSC such as the number of students that requested study groups and number of group created. Created separate components for a new metrics page modeled after the dashboard.

### Test Plan 

Tested using npm run populate to fill the backend with different courses and tested matching different groups and seeing what values were calculated and displayed on the metrics page. Checked that the csvs had the correct values and tested to see that the values changed if different semesters were selected.

Will test more once designs are finalized and Mikayla has met with Jen to confirm numbers are correct

See screenshot below for current state of the page.
Screenshot 2022-11-18 at 3 27 34 PM

### Notes 

Big Work In Progress
- need to clean up code
- need to update design once it is provided

### DIFF
--- a/frontend/src/modules/Metrics/Components/Metrics.tsx
+++ b/frontend/src/modules/Metrics/Components/Metrics.tsx
@@ -1,0 +1,189 @@
+import { useState } from 'react'
+import MenuItem from '@mui/material/MenuItem'
+import { ReactComponent as LogoImg } from '@assets/img/lscicon.svg'
+import { useCourseValue } from '@context/CourseContext'
+import { useStudentValue } from '@context/StudentContext'
+import { AccountMenu } from 'Dashboard/Components/AccountMenu'
+import { Box, Link, SelectChangeEvent } from '@mui/material'
+import { StatGrid } from './StatGrid'
+import { MetricsTable } from './MetricsTable'
+import { DropdownSelect } from '@core/Components'
+
+export const Metrics = () => {
+  const { courses } = useCourseValue()
+  const { students } = useStudentValue()
+
+  //calculate number of unique students who have made requests
+  const num_students = {
+    number: students.length,
+    title: 'unique students',
+    subtitle: 'made requests',
+  }
+
+  //calculate number of unique courses that have received requests
+  const num_courses = {
+    number: courses.length,
+    title: 'courses',
+    subtitle: 'received requests',
+  }
+
+  //calculate total number of requests made by students
+  const num_requests = {
+    number: students.reduce(
+      (total, student) => total + student.groups.length,
+      0
+    ),
+    title: 'requests',
+    subtitle: 'made in total',
+  }
+
+  //calculate number of students matched into groups
+  const num_matches = {
+    number: courses.reduce(
+      (courseTotal, course) =>
+        courseTotal +
+        course.groups.reduce(
+          (groupTotal, group) => groupTotal + group.members.length,
+          0
+        ),
+      0
+    ),
+    title: 'matches',
+    subtitle: 'made in total',
+  }
+
+  //calculate number of students matched into groups
+  const num_groups = {
+    number: courses.reduce((total, course) => total + course.groups.length, 0),
+    title: 'groups',
+    subtitle: 'successfully made',
+  }
+
+  const stats = [
+    num_students,
+    num_courses,
+    num_requests,
+    num_matches,
+    num_groups,
+  ]
+
+  //this can be removed if there is a place to store an objectMap() function
+  const localeMap = (obj: { [key: string]: Date } | undefined) => {
+    if (!obj) {
+      return undefined
+    }
+    return Object.fromEntries(
+      Object.entries(obj).map(([k, v]) => [k, v.toLocaleString()])
+    )
+  }
+
+  const allStudents =
+    courses.length && students.length // Just making sure this isn't calculated until the data is available
+      ? students.flatMap((student) =>
+          student.groups.map((membership) => {
+            const course = courses.find(
+              (c) => c.courseId === membership.courseId
+            )!
+            const group = course.groups.find(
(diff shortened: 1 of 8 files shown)


frontend/src/modules/Metrics/Components/Metrics.tsx
```diff
@@ -0,0 +1,293 @@
...
+    Engineering: 'College of Engineering',
+    HumEc: 'College of Human Ecology',
+    Dyson:
+      'Dyson School of Applied Economics and Management (SC Johnson College of Business)',
+    Nolan:
+      'Nolan School of Hotel Administration (SC Johnson College of Business)',
+    ILR: 'School of Industrial and Labor Relations',
+    SCE: 'School of Continuing Education & Summer Sessions',
+    'Grad School': 'Graduate/Professional School',
+  }
```
mluo24: I think just for consistency you can use a require for the [existing JSON](https://github.com/cornell-dti/zing-lsc/blob/main/frontend/src/modules/Core/Questions/Questions.json) that has this map, like so:

`const questions: Question[] = require('@core/Questions/Questions.json')`

frontend/src/modules/Metrics/Components/StatCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+              background: '#D0EAD6',
+              borderRadius: ' 8px',
+            }}
+          >
+            {' '}
+            <Typography variant="h6" color="#157E2C" fontWeight={600}>
+              + {thisWeek}
+            </Typography>
+          </Box>
```
mluo24: seems like there are some styling issues here (see my pic) so you can either allow some overflow or just fix the width to a better value.

frontend/src/modules/Metrics/Components/Metrics.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+  //this can be removed if there is a place to store an objectMap() function
+  const localeMap = (obj: { [key: string]: Date } | undefined) => {
+    if (!obj) {
+      return undefined
+    }
+    return Object.fromEntries(
+      Object.entries(obj).map(([k, v]) => [k, v.toLocaleString()])
+    )
+  }
```
mluo24, line 117: since this is in AccountMenu, I don't see a reason to repeat this twice.

frontend/src/modules/Metrics/Components/Metrics.tsx
```diff
@@ -0,0 +1,293 @@
...
+          })
+        )
+      : []
+  const getUniqueValues = (arr: any) => {
+    return arr.filter(
+      (value: any, index: any, self: string | any[]) =>
+        self.indexOf(value) === index
+    )
+  }
+  const [selectedRoster, setSelectedRoster] = useState<string>('FA22')
```
mluo24, line 153: it might be worth factoring this out since this is being used in two separate areas (i.e. have a uniform "default semester")

frontend/src/modules/Metrics/Components/Metrics.tsx
```diff
@@ -0,0 +1,293 @@
...
+              groupNumber:
+                membership.groupNumber !== -1
+                  ? `${course.names.join('/')}_${membership.groupNumber}`
+                  : undefined,
+              ...localeMap(group?.templateTimestamps),
+              ...localeMap(membership.templateTimestamps),
+            }
+          })
+        )
+      : []
```
mluo24, line 146: this is also very similar to CSV students. perhaps factoring out also will come in handy?

frontend/src/modules/Metrics/Components/MetricsTable.tsx
```diff
@@ -0,0 +1,90 @@
...
+                <StyledTableCell
+                  sx={{ paddingLeft: '5%' }}
+                  component="th"
+                  scope="row"
+                >
+                  {row.rowName}
+                </StyledTableCell>
+                <StyledTableCell align="center">{row.requests}</StyledTableCell>
+              </StyledTableRow>
+            ))}
```
mluo24, line 74: if there's no data, you can probably display something that says there is no data.

mluo24 (approved):
Thank you Jessica for working on this! I'm blown away by how fast you did this and by how nice it looks! This PR is kind of big, but I mostly paid attention to how you formed the metrics, and by the definitions I think they are implying, they all seem solid. I really like your use of reduce here and your way of handling dates. That being said, there are some nits I have, some of which I addressed individually in the code.

image
There is a visual bug with the "added this week" parts where any 2-digit number will cause it to look wonky.

I have a few questions about design consistency, including the path to metrics being in the dropdown instead of a very visible link and as for what the metrics are (the final designs have four? I'm not sure if all five you have right now are confirmed). Theses are not really present in the designs, but it would also be nice for padding to be at the bottom of the page.

As for code style, you variable name inconsistency with snake_case and camelCase, please just use camelCase. You have types declared at the end in files. You can create a folder called Types and add it to Metrics to store all of the types instead of having it in the files themselves.

That being said, I'm preemptively approving this because it works, and we need to get this merged in very soon. If you could get the fixes done quickly, that would be great!

frontend/src/modules/Metrics/Components/Metrics.tsx
```diff
@@ -0,0 +1,293 @@
...
+    thisWeek: num_students_in_week.length,
+    showAdded: true,
+  }
+  const courses_in_week: string[] = []
+  num_students_in_week.map((member) =>
+    member.map((memberships) => {
+      if (courses_in_week.indexOf(memberships.courseId) === -1) {
+        courses_in_week.push(memberships.courseId)
+      }
+    })
```
mluo24: forgot to mention that you need to fix errors in the console. You don't have anything returning here I think so it's giving an error.

PR #126, opened 2022-11-29
Email template add student to group

rgu0114:
### Summary 

This pull request is the first step towards creating an email template that introduces group members to each other after a new student has been added to an existing group (see [Notion](https://www.notion.so/Email-template-for-add-student-to-group-15de3cf13f714c0ba54986976127680a) for more details)

### Explanation of some important changes 

Adding the new email template: 

I manually added the html file for the new email template, complete with placeholders for names, to the utils folder where all the templates are stored. I added the template metadata to the "add-example-templates.ts" script which gets run each time someone needs to load all the email templates, then exported the emulator data so that it gets added to the emulator/storage_export folder so that we don't need to rerun the script every time. Also made the same additions to Firebase. 
 
Implementing the name substitution support for email template: 

In the Email Modal, we have a function that substitutes values in the email template html so that placeholder values like {{COURSE_NAME}} actually become, say, "CS 2110" in the email. This function takes in (as an argument) a map from strings to strings (e.g. 
    const replaceMap = {
        '{{COURSE_NAME}}': courseName
        *more mappings*...
    })

However, for our new email template that matches a new student with pre-existing groups, we need to replace placeholder names with the actual student names, but we don't have this data available in the original "replacedHtml" function. But we do have the course id and group number (unique identifier) of the group that we're sending an email to. So I created a new function that searches for the correct group, then creates a "replaceMap" that does the following replacement based on the students' order in the group array (newest student will be at the end of the array)
const replaceNamesMap = {
              '{{NEW_STUDENT_NAME}}': newStudent,
              '{{OTHER_STUDENTS_NAMES}}': otherStudents.join(', '),
            }

### Testing 

In instances where multiple groups are selected for this email template, we have a tab for each group that shows its specific template with group members' names, etc.

https://user-images.githubusercontent.com/100158738/218500708-2902788f-adc1-4684-a407-8985ebde2213.mov

https://user-images.githubusercontent.com/100158738/217636881-23ddaddb-f360-401d-a969-35a268d901c9.mov

https://user-images.githubusercontent.com/100158738/205750527-8a6c378e-3822-4cc0-9e69-fddca1d1b680.mov

Notes: 
- The new email template I created is called "Introducing new student to established group" in the popup prompting users to select a template. I have the preview showing the template with unsubstituted fields like {{NEW_STUDENT_NAME}} since at this point, we haven't gone into group-specific emails so we don't have the names yet.
- Couldn't actually fully test that the email went out property as idk if I can send emails in localhost, so instead, I console logged the text of the emails that would've been sent out. As seen in the recording, the template and students show up correctly, but I'm currently figuring out how to make their actual names show up instead of their Cornell emails.


jewang25 (commented):
I like how you followed the formatting of the existing code to create this new template and added a detailed description for every line of code that you added. The screen recording was super helpful and it's nice how you showed what the final email would look like through the console!

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -47,11 +50,57 @@ export const EmailModal = ({
...
+              '{{NEW_STUDENT_NAME}}': newStudent,
+              '{{OTHER_STUDENTS_NAMES}}': otherStudents.join(', '),
+            }
+            console.log(`old students ${otherStudents}`)
+            console.log(
+              Object.entries(replaceNamesMap).reduce(
+                (prev, [key, value]) => prev.replaceAll(key, value),
+                replacedHtml
+              )
+            )
```
mluo24: get rid of `console.logs` once you are done

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -47,11 +50,57 @@ export const EmailModal = ({
...
+   * in the email template. {replacedHtml} is a general replacement common to all groups, so
+   * for this function, we must pass in the specific groupNumber and find the array of
+   * group members' names using that, since this information is not available from {EmailModalProps}
+   *
+   * @param groupNumber
+   */
+  const specificReplacedHtml = async (groupNumber: number) => {
+    const courses = (await axios.get(`${API_ROOT}${COURSE_API}`)).data.map(
+      responseCourseToCourse
+    )
```
mluo24: Don't use axios in this part of the app. These problems can almost always be solved with the courses context or students context, which has already been included in this component (you just need to add the `courses` variable from the context).

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -47,11 +50,57 @@ export const EmailModal = ({
...
   )
 
+  /** Replaces {{NEW_STUDENT_NAME}} and {{OTHER_STUDENTS_NAMES}} with their proper values
+   * in the email template. {replacedHtml} is a general replacement common to all groups, so
+   * for this function, we must pass in the specific groupNumber and find the array of
+   * group members' names using that, since this information is not available from {EmailModalProps}
+   *
+   * @param groupNumber
+   */
+  const specificReplacedHtml = async (groupNumber: number) => {
```
mluo24: I would probably rename this? Ironically, `specificReplacedHtml` is kind of vague.

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -47,11 +50,57 @@ export const EmailModal = ({
...
+    const courses = (await axios.get(`${API_ROOT}${COURSE_API}`)).data.map(
+      responseCourseToCourse
+    )
+
+    courses.forEach((course: Course, index: number) => {
+      if (course.courseId === courseId) {
+        const groups = course.groups
+
+        groups.forEach((group: Group, num: number) => {
+          if (num === groupNumber) {
```
mluo24: off by one error is here

mluo24 (changes requested):
Thank you Richard for working on this task! LSC will now be able to easily send an email easily over for this specific situation. Your general solution works (with adding the template in storage and the template tag + the template tag replacement), but there are two issues with your current implementation of this (that are related to each other): 

The first is that your template replacement doesn't work in the preview, which can cause confusion in that it seems like the template will just send {{NEW_STUDENT_NAME}}. I think this is just a matter of replacing it in Step1 (lines 216-226 in the EmailModal component). It will probably be neater if you combined replaceHtml with your new function so oversights like this won't occur.

The second is that the email you send isn't correct. See the below screenshot for what I mean:
image
The people selected aren't correct, as it's off by one. See my inline comments for the fix. I also think it may be better to use both the student's name and email as opposed to just email. 

Also, get rid of the already added classes to the database. When you load up the emulator, the classes are already there when the database should be empty. This can be solved quickly by exporting without the courses in the database already.

I think testing this out once you make these changes once again will be a good idea!

rgu0114:
> Thank you Richard for working on this task! LSC will now be able to easily send an email easily over for this specific situation. Your general solution works (with adding the template in storage and the template tag + the template tag replacement), but there are two issues with your current implementation of this (that are related to each other):
> 
> The first is that your template replacement doesn't work in the preview, which can cause confusion in that it seems like the template will just send {{NEW_STUDENT_NAME}}. I think this is just a matter of replacing it in Step1 (lines 216-226 in the EmailModal component). It will probably be neater if you combined replaceHtml with your new function so oversights like this won't occur.
> 
> The second is that the email you send isn't correct. See the below screenshot for what I mean: image The people selected aren't correct, as it's off by one. See my inline comments for the fix. I also think it may be better to use both the student's name and email as opposed to just email.
> 
> Also, get rid of the already added classes to the database. When you load up the emulator, the classes are already there when the database should be empty. This can be solved quickly by exporting without the courses in the database already.
> 
> I think testing this out once you make these changes once again will be a good idea!

Thanks for the feedback, will start working on those fixes. In the preview though, the email may be sent to multiple groups. How would the preview look then, since there may be a different NEW_STUDENT_NAME to replace for each group?

frontend/src/modules/EditZing/Components/EmailModal.tsx
rgu0114: Note to self: group 1 has group number (index) 0

Coreball: I haven't been following this PR, but isn't this what `group.groupNumber` would be for, so you aren't relying on `num` (the index) in the groups array? If the index was actually the group number you would be able to access the element directly too.

If these forEach loops only use an if statement to operate on a single element of the array that matches a condition, consider using [Array.find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead to make it more readable and reduce nesting.

For example, I think instead of

```js
courses.forEach((course: Course, index: number) => {
  if (course.courseId === courseId) {
    const groups = course.groups
    groups.forEach((group: Group, num: number) => {
      if (/* condition */) {
        // do something with group
      }
    })
  }
)
```

You could do

```javascript
const group = courses
  .find((course: Course) => course.courseId === courseId)?.groups
  .find((group: Group) => /* condition */)
// do something with group
// group may be undefined if nothing matches
```

Note: I didn't test this and the syntax or semantics may be invalid

mluo24:
> Thanks for the feedback, will start working on those fixes. In the preview though, the email may be sent to multiple groups. How would the preview look then, since there may be a different NEW_STUDENT_NAME to replace for each group?

That's true, @rgu0114, I didn't think about that. I think the simplest option would be to have the preview just show the template, but it's also difficult to tell if you are sending the correct email immediately. I think another solution could be you have different previews for each group when you have this template tag.

frontend/src/modules/EditZing/Components/EmailModal.tsx
rgu0114: Thanks, that works and is much cleaner than before

PR #128, opened 2022-12-08
Fixed Calculating Overall Stats for Metrics Page

jewang25:
### Summary 

Fixed calculating overall stats so that they only show for the chosen semester. Modified IsDateInWeek function so that it calculates the week from beginning of the day Sunday to end of day Saturday.

### Test Plan 
Tested by changing the timestamp for the student submission time for that course on firebase emulator. 

In the screenshot, there are only 8 unique students added for this week because this statistic is calculated based on when the student submitted the first survey request. 

Screenshot 2022-12-08 at 4 44 35 PM
Screenshot 2022-12-08 at 4 43 59 PM
### Breaking Changes 

### DIFF
--- a/frontend/src/modules/Metrics/Components/Metrics.tsx
+++ b/frontend/src/modules/Metrics/Components/Metrics.tsx
@@ -46,10 +46,13 @@ export const Metrics = () => {
   }
 
   const [selectedRoster, setSelectedRoster] = useState<string>('SP23')
+
   const studentsInSemester = new Map<
     string,
     { semester: string; groups: GroupMembership[]; college: string }
   >()
+
+  //calculates details for all students
   const allStudents =
     courses.length && students.length // Just making sure this isn't calculated until the data is available
       ? students.flatMap((student) =>
@@ -86,6 +89,7 @@ export const Metrics = () => {
           })
         )
       : []
+
   const getUniqueValues = (arr: any) => {
     return arr.filter(
       (value: any, index: any, self: string | any[]) =>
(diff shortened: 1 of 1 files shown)


JustKong13 (commented):
Very nice, quick change to provide more accurate metrics on the recent number of students joining! Good work!

frontend/src/modules/Metrics/Components/Metrics.tsx
```diff
@@ -36,12 +36,18 @@ export const Metrics = () => {
       isDateInThisWeek(membership.submissionTime)
     )
   )
+  const uniqueStudentsInWeek = []
+  for (let i = 0; i < numStudentsInWeek.length; i++) {
+    if (numStudentsInWeek[i].length === students[i].groups.length) {
+      uniqueStudentsInWeek.push(students[i])
+    }
+  }
```
CollinWoo: I think that this section could be simplified with the use of the `filter()` function:
`const uniqueStudentsInWeek = numStudentsInWeek.filter((student, i) => student.length === students[i].groups.length);`

The only catch to this method is that you get an array of type `GroupMembership[] ` instead of `Student`. If you really do need the array to contain the actual students, using `forEach()` on the `numStudentsInWeek` array will allow you to do the same thing you did here but with less code:

`const uniqueStudentsInWeek = []`
`numStudentsInWeek.forEach((student, i) => {
    if (student.length === students[i].groups.length) {
      uniqueStudentsInWeek.push(students[i])
    }
  })`

b3njamint:
Nice work! Really good job on the metrics page, and this is a good fix too with a notable impact on the users. Collin's point about using .filter makes sense and you should implement it, but it is very minor.

b3njamint (commented):
Nice work! Really good job on the metrics page, and this is a good fix too with a notable impact on the users. Collin's point about using .filter makes sense and you should implement it, but it is very minor.

frontend/src/modules/Metrics/Components/Metrics.tsx
mluo24: As Colin said, `forEach` is a good idea here if you're trying to iterate between these two groups that are the same size. Same for other places where you have a for loop (which I believe exist since I didn't mention this last time).

mluo24 (changes requested):
Thanks Jessica for getting this done so quickly! I think others have touched upon the code quality observations already. I think the functionality works, except for one caveat:

This isn't in this PR specifically, but in your `isDateInThisWeek` function, you don't change the time of the `firstDayOfWeek` variable to be 12 AM, so if your timestamp is before that time on a Sunday, it will register as being being last week. The same goes for the last day of the week, where it should extend until the end of the day. This doesn't mean your code is wrong (in fact, I think it is technically more "correct" since that is exactly the range it should be since you query your current day), but you generally consider "this week" to mean the beginning of the day Sunday to end of day Saturday.

One way to fix this is to add these few lines appropriately, some of which you already have:
```
// set the date to be Sunday midnight
todayObj.setDate(todayDate - todayDay) 
todayObj.setHours(0, 0, 0, 0)

// ...

// set the date to be Saturday 11:59:59 PM
lastDayOfWeek.setDate(lastDayOfWeek.getDate() + 6)
lastDayOfWeek.setHours(23, 59, 999)
```

Great work!

frontend/src/modules/Metrics/Components/Metrics.tsx
```diff
@@ -10,79 +10,171 @@ import { MetricsTable } from './MetricsTable'
...
+      studentsInSemester.get(student) != null
+        ? studentsInSemester.get(student)?.groups.length
+        : 0
+    if (studentValues.groups.length === originalGroups) {
+      uniqueStudentsInWeek += 1
+    }
+    studentValues.groups.forEach((group) => {
+      if (coursesInWeek.indexOf(group.courseId) === -1) {
+        coursesInWeek.push(group.courseId)
+      }
```
mluo24: instead of this, you can use a Set which should guarantee unique values

frontend/src/modules/Metrics/Components/Metrics.tsx
```diff
@@ -10,79 +10,171 @@ import { MetricsTable } from './MetricsTable'
...
+          groups: studentValues.groups.filter((group) =>
+            isDateInThisWeek(group.submissionTime)
+          ),
+          college: studentValues.college,
+        })
+      }
+      return (
+        isDateInThisWeek(membership.submissionTime) &&
+        numStudentsInWeek.get(student) == null
+      )
```
mluo24: I am not quite sure what this map function is for, especially since its result doesn't have its value assigned to any variable. You can just iterate through the groups with a forEach again.

frontend/src/modules/Metrics/Components/Metrics.tsx
```diff
@@ -10,79 +10,171 @@ import { MetricsTable } from './MetricsTable'
...
+        isDateInThisWeek(membership.submissionTime) &&
+        numStudentsInWeek.get(student) == null
+      ) {
+        numStudentsInWeek.set(student, {
+          semester: studentValues.semester,
+          groups: studentValues.groups.filter((group) =>
+            isDateInThisWeek(group.submissionTime)
+          ),
+          college: studentValues.college,
+        })
```
mluo24: What's the purpose of making this a map instead of another array? I feel like the main information is the number of groups that are within this week so you can compare it later on. Or you can do the comparison in here directly instead of doing it within the forEach of numStudentsInWeek.

frontend/src/modules/Metrics/Components/Metrics.tsx
```diff
@@ -10,79 +10,171 @@ import { MetricsTable } from './MetricsTable'
...
+            const group = course.groups.find(
+              // undefined if student is unmatched
+              (g) => g.groupNumber === membership.groupNumber
+            )
+            if (course.roster === selectedRoster) {
+              studentsInSemester.set(student.email, {
+                semester: course.roster,
+                groups: student.groups,
+                college: student.college,
+              })
```
mluo24, line 152: Is there a reason for this structure?

mluo24 (approved):
Thank you for making these fixes so promptly! From what I can tell, everything seems to be in working order, and the students are now just those who have existed in the current semester in addition to the courses, though I'm a little confused by some of the design choices in the new code, which I have left comments on. However, the filtering of students by semester, the logic of "this week," and unique students this week all look good to me! I think it's also worth leaving a comment somewhere (notion, in the code, etc.) on the criteria of being a unique student in that week (a student who has not submitted a survey before who has submitted a survey in the past week).

I'm approving this since it works, but I appreciate addressing my comments. Thanks again!

mluo24 (commented):
btw merge in main and resolve conflicts so I can look at this again, but this should be fine since we've talked last time

PR #132, opened 2022-12-09
Undo delete

JustKong13:
### Summary 

This pull request is to implement the ability to undo the deletion of a group if a deletion was done by accident. If the trash can on a group was clicked, it would display a button to confirm the deletion of the group, or the ability to undo the delete. 

- [x] implemented functions to undo a delete
- [x] Completion of the front end functionality 

### Test Plan 

image

### DIFF
--- a/backend/functions/src/matching/functions.ts
+++ b/backend/functions/src/matching/functions.ts
@@ -322,33 +322,23 @@ async function createEmptyGroup(courseId: string) {
   await Promise.all([groupCreationUpdate, groupNumberUpdate])
 }
 
-async function hideEmptyGroup(courseId: string, groupNumber: number) {
+async function hideEmptyGroup(
+  courseId: string,
+  groupNumber: number,
+  toHide: boolean
+) {
   await assertIsExistingCourse(courseId)
 
   const groupToHide = courseRef
     .doc(courseId)
     .collection('groups')
     .doc(groupNumber.toString())
-    .update({ hidden: true })
+    .update({ hidden: toHide })
     .catch((err) => {
       console.log(err)
       throw new Error(`Error in removing group ${groupNumber}`)
     })
 
-  await Promise.all([groupToHide])
-}
-
-async function unhideGroup(courseId: string, groupNumber: number) {
-  await assertIsExistingCourse(courseId)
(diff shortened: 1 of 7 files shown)


backend/functions/src/matching/functions.ts
```diff
@@ -338,6 +338,20 @@ async function hideEmptyGroup(courseId: string, groupNumber: number) {
...
+    .collection('groups')
+    .doc(groupNumber.toString())
+    .update({ hidden: false })
+    .catch((err) => {
+      console.log(err)
+      throw new Error(`Cannot show group ${groupNumber}`)
+    })
+  await Promise.all([groupToHide])
+}
+
```
mluo24: I think you can change the function to be more general, in that you just pass in `true` or `false` depending on the use case. The endpoint can remain separate, but this function used to implement those endpoints can be more general.

backend/functions/src/matching/routes.ts
```diff
@@ -106,4 +107,14 @@ router.post('/hide-group', (req, res) => {
     })
 })
 
+router.post('/unhide-group', (req, res) => {
+  const { courseId, groupNumber } = req.body
+  unhideGroup(courseId, groupNumber)
```
mluo24: This line (112) is where you can replace with a more general function, and also in the other endpoint.

mluo24 (approved):
Thanks for working on this fix Justin! This provides an extra layer of protection if LSC didn't mean to delete a group, which is helpful!

Your current implementation isn't exactly what I had in mind, but there are no designs for this as of now, so I will approve it. I think it is slightly interesting to delete it and then restore it at the confirmation stage, as semantically, "confirming" generally indicates it will not be deleted until you click "confirm." I was more thinking of a popup at the bottom that says "undo" if you want to delete (which is similar to if you delete an event on Google Calendar). 

But besides my comments, everything seems to be working and looks good! Nice work, I will be approving.

JustKong13:
The suggestion to create a more general function that takes in an additional boolean was a good one to clean up the code and not have redundant functions. I've kept the separate endpoints and the two separate calls in `GroupCard.tsx`, but everywhere else should use the same function now.

PR #133, opened 2022-12-09
New filter: Can Add to Existing Group

rgu0114:
### Summary 
Created a new filter for classes called "Can Add to Existing Group." This filter is for when we have at least one existing group and exactly one unmatched student, so we can add them to an already existing group. 

### Test Plan 

As seen below, the new filter shows up correctly and contains all classes that fit the above description.

https://user-images.githubusercontent.com/100158738/206579675-52623fcf-aad2-4474-8107-fabe428fcaa6.mov

### DIFF
--- a/frontend/src/modules/Dashboard/Components/Dashboard.tsx
+++ b/frontend/src/modules/Dashboard/Components/Dashboard.tsx
@@ -22,6 +22,7 @@ type FilterOption =
   | 'unmatchable'
   | 'newly-matchable'
   | 'matchable'
+  | 'can-add-to-existing-group'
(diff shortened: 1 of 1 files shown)


mluo24 (approved):
Thanks for getting this PR and feature together so quickly after the misunderstanding (sorry about that!). Your solution is clean, and I also really like the positioning of it in the filter dropdown. Everything looks good and it works so I'm approving it.

Also make sure to delete this branch remotely when you merge it in because "testing" is too general of a name (and also does not match this branch) and may be reserved for other things in the future. Also probably delete the old remote branches that you made to reduce clutter.

PR #136, opened 2022-12-10
Loading for signing in

b3njamint:
### Summary 

This pull request changes the admin login button to visually show the user that their login request is being handled by using a LoadingButton.

- [x] implemented X
- [ ] fixed Y

### Test Plan 

Before clicking signin button.
Screen Shot 2022-12-09 at 8 57 36 PM

While signin button is clicked and the request is being handled. There is a little spinny thing.
Screen Shot 2022-12-09 at 8 57 40 PM

### Notes 

I changed the button color when loading to white from grey, to be more in line with the theme of Zing. I don't think there is a loading button in the design but correct me if I am wrong.

### Breaking Changes 

Nothing.

### DIFF
--- a/frontend/src/modules/AdminHome/Components/AdminHome.tsx
+++ b/frontend/src/modules/AdminHome/Components/AdminHome.tsx
@@ -78,7 +78,9 @@ export const AdminHome = () => {
           }}
           onClick={() => {
             handleSigninClick()
-            adminSignIn().catch(() => {})
+            adminSignIn().catch(() => {
+              setLoading(false)
+            })
           }}
         >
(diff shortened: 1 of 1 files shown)


mluo24 (approved):
Hi Benji, thanks for working on this small but helpful fix! This should give enough for user feedback to know that the signing in is occurring. Nice use of the loading button! 

I have a small nit for catching if the signin goes wrong, which I have noted in the code. That line should set the loading to false should something go wrong with signing in. Theoretically, in the future, this should also show some error message, but I think the loading stopping is an indicator that the request is done and nothing happened.

PR #138, opened 2022-12-21
Change roster to WI23

mluo24:
### Summary 

Change the roster for classes that the survey uses to use Winter 2023 instead of Fall 2022. When students submit surveys now they will be added to the Winter 2023 iteration of the course.

- [x] change roster to WI23 (both where the surveys go in the backend and default in dashboard)
- [x] changed already present option Winter 2022 to Winter 2023

### Test Plan 

I ran the `npm run populate` script and saw if everything was Winter 2023 (correct roster, so not all courses would be added, and dashboard would show the same, and the URL will show the same).


Coreball (approved):
☃️

PR #139, opened 2022-12-31
Migrate from CRA to Vite

mluo24:
### Summary 

Create React App (referred to as "CRA," which is what the React app was originally initialized with) is generally slow (build times take forever when running the dev server) among other issues, and Vite (another frontend framework bundler that does a similar thing to CRA) has a lot of advantages over CRA currently that there is no reason to keep it. See [this](https://semaphoreci.com/blog/vite) if you're not compelled. See [this](https://cathalmacdonnacha.com/migrating-from-create-react-app-cra-to-vite) for the steps followed for the migration. 

- [x] installed Vite
- [x] refactored the frontend structure slightly
- [x] replaced package.json with Vite scripts and customization of Vite
- [x] removed CRA

### Test Plan 

https://user-images.githubusercontent.com/69130822/210126585-b378279c-9ec8-4322-bf55-3f63f8d53fdc.mov

(it's so fast)

as of right now, no deployment has been tried. 

### Notes 

I don't think there are any, only the build system changed. Make sure to run `yarn` on the frontend to get the latest changes. I also removed some of the other scripts that were reliant on CRA and therefore no longer work (so don't use those).

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -2,34 +2,24 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
-    <link
-      rel="icon"
-      type="image/png"
-      sizes="32x32"
-      href="%PUBLIC_URL%/favicon-32x32.png"
-    />
-    <link
-      rel="icon"
-      type="image/png"
-      sizes="16x16"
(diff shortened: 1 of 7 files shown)


wizhaaa (approved):
Awesome!! Vite is very fast indeed. wow so speed. much fast. Def much faster than CRA.

PR #142, opened 2023-02-08
No match, session closed :(

b3njamint:
### Summary 

This pull request adds an email template for students who do not have a match after about a 6-week period. Poor students :(

I made and added the new template HTML file, then ran some scripts to add the template to the backend/Firebase. With advice from template master @rgu0114!

- [x] implemented X
- [ ] fixed Y

### Test Plan 

bt283 seems to have been there for 6 weeks now. It's time to send him an email.
Screen Shot 2023-02-08 at 3 26 47 PM

The email template is listed!
Screen Shot 2023-02-08 at 3 27 00 PM

As you can see, {{COURSE_NAME}} is replaced with the course name, PHYS 2213. The hyperlinks are correct too.
Screen Shot 2023-02-08 at 3 27 07 PM

You can also modify the template and it works too. Note the "Hello" -> "Hello Zinger."
Screen Shot 2023-02-08 at 3 27 45 PM

### Notes 

This email should send just like any other template, but I haven't tried that as I don't have email-sending permissions.

### Breaking Changes 

I added a new HTML file in gs://zing-lsc-templates called `no-match-session-closed.html`


JustKong13 (commented):
This PR looks lengthy, but most of it looks like simple HTML revising, and revising emails. I understand that you don't have emailing permissions so it's difficult to test that this function works but nothing seems to break. Well done! And thanks for taking on this difficult task

Coreball:
> This PR looks lengthy

It seems to me that the emulator files were not always exported in the proper location and those files were committed instead of deleted. For example the path `backend/functions/src/utils/emulator/` is non-standard but contains files in this pull request

wizhaaa:
^ Mentioned during the work session, but it does look like you have some changes from some other branch (ie. the emailing files). Probably an error in rebasing.

rgu0114 (commented):
Nice work on this PR Benji! Good job on the email templates – you figured that out pretty quickly and it looks like the correct template is being generated. However, I was curious how you keep track of how long a student has been unmatched for. Is that something that's still done manually, or an indicator or something that still needs to be added?

backend/functions/src/utils/no-match-session-closed.html
```diff
@@ -0,0 +1,12 @@
+<p>Hello,</p>
```
wizhaaa, line 1: Comment to self: this is the no match session closed HTML file.

wizhaaa:
Just a bump to formally document progress: 

- [ ]  I think you still have some residual code (ie. files have been changed in the emailing functions with leftover console logging. 
- [ ] The emailing uploading directions can be found here: https://www.notion.so/How-to-Add-Email-Templates-b99a786b50924ec4bb41332bde3f4a60?pvs=4

PR #144, opened 2023-02-08
Fixed requires with Object.entries instead

jewang25:
### Summary 

Fixed bug with survey because requires does not work with vite. Used object.entries instead to interpret the json values so that it could be displayed on the frontend. This affected both the survey and metrics page.

### Test Plan 

Tested walking through survey and checking metrics page when populated with classes to see if the right college names showed up.
Screenshot 2023-02-12 at 5 48 33 PM
Screenshot 2023-02-12 at 5 48 02 PM


mluo24 (approved):
Nice work Jessica and thanks for working on this! After the Vite change, `require` doesn't work anymore because I think it was only a webpack thing. So, we had to change the `require`s to be something completely different. Here is a list of references for the issue:

https://vitejs.dev/guide/features.html#json
https://stackoverflow.com/questions/12710905/how-do-i-dynamically-assign-properties-to-an-object-in-typescript
https://github.com/vitejs/vite/issues/3409

PR #145, opened 2023-02-10
Settings page

JustKong13:
### Summary 

This pull request implements the frontend for the settings page. There is an option to change the semesters, and more as needed. 

This pull request is the first step towards implementing feature Foo

- [x]  Create new component for settings
- [x]  Add page to list of routes in react router

### Test Plan 
No test plan because this was a frontend task, but here is an image of the page!!
image

### DIFF
--- a/frontend/src/modules/Core/Constants/Routes.ts
+++ b/frontend/src/modules/Core/Constants/Routes.ts
@@ -7,6 +7,7 @@ export const DASHBOARD_PATH = '/dashboard'
 export const METRICS_PATH = '/metrics'
 export const EMAIL_PATH = '/email'
 export const TEMPLATE_EDITOR_PATH = '/templates'
+export const SETTINGS_PATH = '/dashboard/settings'
 
 // NODE_ENV will be 'production' for yarn build, and 'development' for yarn start
 // See https://create-react-app.dev/docs/adding-custom-environment-variables/
--- a/frontend/src/modules/Core/Constants/Routes.ts
+++ b/frontend/src/modules/Core/Constants/Routes.ts
@@ -7,7 +7,7 @@ export const DASHBOARD_PATH = '/dashboard'
 export const METRICS_PATH = '/metrics'
 export const EMAIL_PATH = '/email'
 export const TEMPLATE_EDITOR_PATH = '/templates'
-export const SETTINGS_PATH = '/dashboard/settings'
+export const SETTINGS_PATH = '/settings'
 
 // NODE_ENV will be 'production' for yarn build, and 'development' for yarn start
 // See https://create-react-app.dev/docs/adding-custom-environment-variables/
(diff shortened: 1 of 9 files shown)


frontend/src/modules/Settings/Components/Settings.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+        <DropdownSelect
+          value={currRoster}
+          onChange={changeCurrRoster}
+          sx={{
+            alignContent: 'right',
+            right: '1px',
+          }}
+        >
+          <MenuItem value="SU22">Summer 2022</MenuItem>
```
wizhaaa: So, I believe Colin's PR will provide a backend endpoint to get the current and list of available semesters - the goal being that the semester "state" is stored in the backend and is global for all users. One thing you could do to to prepare for it is to create a list for now (of the different semesters { value and name } ) and then map through that list to render `MenuItems`.

wizhaaa (approved):
I think this is about the most you can do right now until Colin's PR is finished and gives access to the backend endpoints + functions actually to update the state. 

One thing I'm not positive about is if we still want the "Switch Semester" in the menu if we add it to the Settings Page - I personally think it may be redundant (but it is one less click, but since we want to _globally_ switch semesters for all users, it should only be accessed around ~1 a semester so it doesn't have to be easily accessible). This may be another PR or something to consider in the future after discussing it with PM/designers.

frontend/src/modules/Core/Constants/Routes.ts
```diff
@@ -7,6 +7,7 @@ export const DASHBOARD_PATH = '/dashboard'
 export const METRICS_PATH = '/metrics'
 export const EMAIL_PATH = '/email'
 export const TEMPLATE_EDITOR_PATH = '/templates'
+export const SETTINGS_PATH = '/dashboard/settings'
```
mluo24: Since there haven't been any nested paths, I think we can keep the trend of just using `/settings` instead of having it nested in dashboard

frontend/src/modules/Dashboard/Components/AccountMenu.tsx
```diff
@@ -144,6 +144,11 @@ export const AccountMenu = ({
             Dashboard
           </MenuItem>
         )}
+        {showSettingsLink && (
+          <MenuItem component={Link} to="/dashboard/settings">
+            Settings
+          </MenuItem>
+        )}
```
mluo24: since this isn't in use yet i don't think you need to worry about putting it into the menu just yet (in case we deploy so we don't confuse the LSC staff).

frontend/src/modules/Settings/Components/Settings.tsx
mluo24: ^ yeah eventually this is will be programmatic and not hardcoded so you can make this into some data array and render it in a map rather than hard coding it here now.

mluo24 (approved):
lgmt, one small nit is that you could also put settings as part of dashboard (like how metrics is right now) since they are loosely related. Otherwise, good first direction into this feature! Thanks Justin :)

PR #146, opened 2023-02-11
Add Student to Existing Group

rgu0114:
### Summary 

We created a new email template for the situation where we want to add a student to an existing group and notify both parties. This PR consists of two steps:

1) Creating the email template itself and ensuring that the appropriate name substitutions are made. 
2) Ensuring that the group-specific emails are easily visible in email preview

### Test Plan 

Below, we run through the user flow of sending an email to group(s) that have just had a new member added to an existing group:

https://user-images.githubusercontent.com/100158738/218235232-73810e5b-47e5-44cc-be49-5f2d4c02482a.mov

Email now contains students' actual names instead of just their emails! 

Screen Shot 2023-02-14 at 11 50 43 AM


jewang25 (commented):
Nice job mapping everything so that a separate email shows up for each group. I think you could get rid of the console.log comments to clean up the code a little. Not sure why so many backend files were modified, but other than that good job!

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -48,12 +51,53 @@ export const EmailModal = ({
...
+      '{{NEW_STUDENT_NAME}}': newStudent,
+      '{{OTHER_STUDENTS_NAMES}}': otherStudents.join(', '),
+    }
+    console.log(`old students ${otherStudents}`)
+    console.log(
+      Object.entries(replaceNamesMap).reduce(
+        (prev, [key, value]) => prev.replaceAll(key, value),
+        replacedHtml
+      )
+    )
```
mluo24: remove console.logs

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -93,18 +182,31 @@ export const EmailModal = ({
...
-        }
-        return sendEmail(emailItems)
-      })
+      selectedTemplate.name === 'Introducing student to established group'
+        ? allGroupTemplates.map(({ groupNumber, template }) => {
+            const emailSubject = selectedTemplate?.subject
+            const emailItems = {
+              emailSubject,
+              indivEmail: undefined,
+              emailBody: template,
```
mluo24: This line is incorrect (`emailBody: template`). Change it to be the same as line 203.

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -163,7 +265,7 @@ export const EmailModal = ({
           variant="text"
           sx={{ alignSelf: 'end' }}
         >
-          <EditIcon sx={{ marginRight: '5px' }} />
+          <EditIcon style={{ marginRight: '5px' }} />
```
mluo24: For what reason was this change?

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -192,27 +316,77 @@ export const EmailModal = ({
...
+                  fontSize: 24,
+                  fontWeight: 900,
+                  maxWidth: '90%',
+                  whiteSpace: 'nowrap',
+                }}
+              >
+                Group {groupNumber}
+              </Box>
+
+              <EmailPreview template={template} replacedHtml={template.html} />
```
mluo24: This line causes it to not show the replaced HTML. I think you should change the function to also include a template you can pass into it.

mluo24 (commented):
You are almost there Richard! I'm not approving yet because there are some big bugs that need to be squashed before merging which can be resolved very quickly. I pointed the specific places out in code but here they are:

1. in the main preview with all the groups, you don't show the replaced HTML with the template tags, but you have the correct output in the all groups editor! 

2. Email sending doesn't work, which is a big one caused by having the wrong email body.

Otherwise, great work on this! Thank you for being willing to spend months on this task.

frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -40,20 +49,100 @@ export const EmailModal = ({
...
+    if (!group) {
+      return replacedHtml
+    }
+
+    const groupSize = group.members.length
+
+    const members = group.members.map(
+      (member) =>
+        students.find((student) => student.email === member)?.name || ''
+    )
```
mluo24, line 87: Also I forgot to remind you to change this to the students' actual names rather than netids.

frontend/src/modules/EditZing/Components/EmailModal.tsx
rgu0114: idr when I changed this lol, I'll change it back

mluo24 (approved):
nice richard

rgu0114:
> nice richard

thank

PR #148, opened 2023-02-12
Add API Endpoints for Switching Semesters

CollinWoo:
### Summary 
This pull request adds API endpoints to allow for semesters to be switched manually. These endpoints include /course/semester/current (to set and retrieve the current semester) and /course/semester/all (to retrieve all semesters).

- [x] Added API endpoints to retrieve and set the current and past semesters
- [x] Modified Firestore to store the current and past semesters

### Test Plan 
Testing was conducted in Insomnia by making requests to the appropriate endpoints and verifying that the returned data is correct.

POST request to /course/semester/current
Screenshot 2023-02-12 at 11 41 36 AM

GET request to /course/semester/current
Screenshot 2023-02-12 at 11 41 48 AM

GET request to /course/semester/all
Screenshot 2023-02-12 at 11 43 17 AM

### Breaking Changes 

- Added a new utils collection to the Firestore which contains a semesters doc that stores current and past semesters

### DIFF
--- a/backend/functions/src/course/functions.ts
+++ b/backend/functions/src/course/functions.ts
@@ -16,7 +16,7 @@ export const setCurrentSemester = async (sem: String) => {
 
 export const getAllSemesters = async () => {
   const semData = (await semesterRef.get()).data() as Semester
-  return semData.allSemesters
+  return [semData.currentSemester, ...semData.allSemesters]
 }
 
 async function getCourseInfo(courseId: string) {
--- a/backend/functions/src/course/functions.ts
+++ b/backend/functions/src/course/functions.ts
@@ -1,8 +1,23 @@
 import { db } from '../config'
 import { getStudentsData } from '../student/functions'
 import admin from 'firebase-admin'
-import { Course, FirestoreCourse, FirestoreGroup } from '../types'
+import { Course, FirestoreCourse, FirestoreGroup, Semester } from '../types'
 const courseRef = db.collection('courses')
+const semesterRef = db.collection('utils').doc('semester')
+
+export const getCurrentSemester = async (): Promise<String> => {
+  const semData = (await semesterRef.get()).data() as Semester
+  return semData.currentSemester
+}
+
+export const setCurrentSemester = async (sem: String) => {
+  return semesterRef.set({ currentSemester: sem })
+}
+
+export const getAllSemesters = async () => {
+  const semData = (await semesterRef.get()).data() as Semester
+  return semData.allSemesters
+}
 
 async function getCourseInfo(courseId: string) {
   const snapshot = await courseRef.doc(courseId).get()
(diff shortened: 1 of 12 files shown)


backend/functions/src/course/routes.ts
```diff
@@ -35,4 +42,32 @@ router.get('/students/:courseId', (req, res) => {
...
+      res.status(500).send({ message: err.message })
+    })
+})
+
+router.post('/semester/current', (req, res) => {
+  const { semester } = req.body
+  setCurrentSemester(semester)
+    .then(() => res.status(200).send({ success: true }))
+    .catch((err) => {
+      console.log(err)
```
wizhaaa: Should remove the `console.log` here

backend/functions/src/course/routes.ts
```diff
@@ -35,4 +42,32 @@ router.get('/students/:courseId', (req, res) => {
     })
 })
 
+router.get('/semester/current', (_, res) => {
+  getCurrentSemester()
+    .then((data) => res.status(200).send(data))
+    .catch((err) => {
+      logger.error(`Unexpected error getting semester: ${err.message}`)
```
wizhaaa: I think one thing you could do to make it a little refactored/useful is set up error logging like so: 

```javascript
.catch( (err) => { 
	const err_msg = `Unexpected error getting semester: ${err.msg}` 
	logger.error(err_msg); 
	res.status(500).send({ message: err_msg })
})	

```

Can also change the error message for the other endpoints to be like: 
```javascript 
`Error changing current semester to ${semester}`
```

So that the error message tells us more of what the user input is.

wizhaaa (approved):
Thanks for making the endpoints! 

I think it mainly looks fine, there are possibly some quality-of-life error messaging improvements you could make to make the error + debugging easier. 

The next steps would be to actually change the actual semester for the dashboard - which is probably not in scope of this PR.

backend/functions/src/course/functions.ts
```diff
@@ -1,8 +1,23 @@
...
+import { Course, FirestoreCourse, FirestoreGroup, Semester } from '../types'
 const courseRef = db.collection('courses')
+const semesterRef = db.collection('utils').doc('semester')
+
+export const getCurrentSemester = async (): Promise<String> => {
+  const semData = (await semesterRef.get()).data() as Semester
+  return semData.currentSemester
+}
+
+export const setCurrentSemester = async (sem: String) => {
```
JustKong13, line 13: For `setCurrentSemester`, I think it may be necessary to check, and add the semester to `allSemesters` just in case we are adding a new semester. Apart from that the endpoints looks good!

JustKong13 (commented):
The endpoints all look good, and the new type semester you added is also short, easy to understand. 

I just have one concern about the setting semester endpoint, I'm not sure whether `setCurrentSemester` will be able to add a new semester that isn't already in `allSemesters`, or whether we are planning on having a different endpoint do that for us.

PR #149, opened 2023-02-15
Add .csv extension to downloads

b3njamint:
### Summary 

This pull request fixes the bug in Safari where downloaded `.csv` files are not recognized.

From a little bit of research, Safari does not infer the file type of the download so without further specification, you just get a plain text file. Chrome, on the other hand, is superior and will do so.

Therefore, all I needed to do was specify the file type in defining the file name.

- [] implemented X
- [x] fixed Y

### Test Plan 
Here are the downloads from Safari. As you can see, they are now `.csv` files!
Screen Shot 2023-02-15 at 4 43 58 PM

Here are the downloads from Chrome. Since Chrome is smart, it does not become `filename.csv.csv`.
Screen Shot 2023-02-15 at 4 44 05 PM

### Notes 

I have not tested this with other browsers, but I think most browsers either infer **missing** file types, or don't infer at all, so `filename.csv.csv` should not happen.

### Breaking Changes


wizhaaa (approved):
LGTM, works well for me on safari too. 

One thing we could do to improve this is to have a better, more informative file name format.  

One format I'm thinking is: "LSC Exported Courses Year-Month-Date" using `Date.getYear()` etc.

b3njamint:
> LGTM, works well for me on safari too.
> 
> One thing we could do to improve this is to have a better, more informative file name format.
> 
> One format I'm thinking is: "LSC Exported Courses Year-Month-Date" using `Date.getYear()` etc.

Thanks! I changed the format of the downloads accordingly to "lsc-exported-courses-YYYY-MM-DD.csv"

Coreball:
> One thing we could do to improve this is to have a better, more informative file name format.

Would be useful to ask them directly what they would like

PR #150, opened 2023-02-20
Changed default to be none when picking email templates

jewang25:
- added undefined as value for selected template

### Summary 

- The next button is disabled until a template is selected. If the back button is clicked, the previously chosen template stays selected. 

### Test Plan 

I went back and forth on the pages and tested the email template modal for various different groups.

https://user-images.githubusercontent.com/94809605/219997663-b4c4e4d3-5e07-4132-a685-40bdc50b1682.mov

### DIFF
--- a/frontend/src/modules/Core/Types/FormFieldProps.ts
+++ b/frontend/src/modules/Core/Types/FormFieldProps.ts
@@ -57,7 +57,7 @@ export interface GoToButtonProps extends Common {
 }
 
 export interface RadioButtonProps extends Common {
-  currentAnswer: string
+  currentAnswer: string | undefined
   onClick: (e: React.ChangeEvent<HTMLInputElement>) => void
   value: string
   label: string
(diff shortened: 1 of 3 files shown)


frontend/src/modules/EditZing/Components/EmailModal.tsx
```diff
@@ -301,7 +301,7 @@ export const EmailModal = ({
           onClick={() => {
             setStep(1)
           }}
-          disabled={!selectedTemplate}
+          disabled={!selectedTemplate || selectedTemplate === undefined}
```
mluo24: this logic is redundant (!selectedTemplate already covers the undefined case)

frontend/src/modules/Core/Types/FormFieldProps.ts
```diff
@@ -57,7 +57,7 @@ export interface GoToButtonProps extends Common {
 }
 
 export interface RadioButtonProps extends Common {
-  currentAnswer: string
+  currentAnswer: string | undefined
```
mluo24: I don't think this is needed

mluo24 (approved):
Thank you Jessica for taking care of this important, high priority task, and nice job on it! I think you covered the edge cases well and everything works as intended. See my comments (minor changes) and this should be good to go :)

I hope you are feeling better!

PR #154, opened 2023-03-03
Fix errors in metrics revert

mluo24:
### Summary 

See #128

image

I did the last thing.

I tested it and it works!


wizhaaa (approved):
Looks like you managed to fix it - nice work! Must have been tedious.

PR #155, opened 2023-03-03
Fixed matchable filter

rgu0114:
### Summary 

We have a bug where the "matchable" filter contains classes where there's only one unmatched student with nonzero groups, but this case should instead be contained only within the "can add to already existing group" filter. 

For example, in the below filter instance, ILRST 2100 should not appear in matchable.
Untitled

To fix this, we just ensure that the "matchable" criteria doesn't include the case where there's exactly one unmatched student and multiple available groups.

### Test Plan 

CS 2110 has nonzero groups and one unmatched student, but it no longer appears in "matchable" and instead appears only in "can add to already existing group" filter.

https://user-images.githubusercontent.com/100158738/222845295-70ee0896-d36f-402c-a584-0c717627382d.mov

### DIFF
--- a/frontend/src/modules/Dashboard/Components/Dashboard.tsx
+++ b/frontend/src/modules/Dashboard/Components/Dashboard.tsx
@@ -104,7 +104,7 @@ export const Dashboard = () => {
       case 'matchable':
         return [...courseInfo].filter(
           (course, _) =>
-            (course.lastGroupNumber > 0 && course.unmatched.length > 0) ||
+            (course.lastGroupNumber > 0 && course.unmatched.length > 1) ||
             (course.lastGroupNumber === 0 && course.unmatched.length > 1)
         )
       case 'can-add-to-existing-group':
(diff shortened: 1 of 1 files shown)


CollinWoo:
Great job Richard! I love how simple but effective this fix is. My only concern (which is kinda unrelated to the matchable filter) is that the "can add to already existing group" filter doesn't account for the possibility that there is one new student but all existing groups are full. In this case, I'm don't think this new student can be added to any existing group, but the class would still show up under "can add to already existing group".

jewang25 (commented):
I agree with Colin that it's cool how this was such a simple fix. It looks like it all works so good job! In terms of what Colin mentioned about the can add to an existing group filter, it makes sense to consider group limit, but I'm not sure if we consider it in the rest of our code so it might not be needed.

wizhaaa (approved):
A clever solution here Richard! Good work. Simple but works as far as I can tell.

To address Colin's concern regarding the `'can-add-to-existing-group'` filter even when groups are full - I don't believe there is a maximum group member as of now. So should be fine - especially if they want to just add an additional member later in the semester and there is a low chance of more students submitting surveys (the only option would be to join a pre-existing group regardless of its size).

PR #157, opened 2023-03-06
PR: Serial Number

jewang25:
### Summary

Added a unique groupId to every group created using the unique courseId and group number.

### Test Plan 

- I plan to test by logging errors and matching groups in multiple courses. I will check the value of the groupId in firebase as well as the value of the counter and compare it with the value displayed on the frontend to make sure the feature is working. I will also look for the groupId in the csv to see if it is correct and can be used to find the group on the website. I will also test all the cases such as transferring students and deleting groups to see if the value updates correctly.

### Breaking Changes 

- changes the firebase collection store to add groupId to each group
- changes group object type in the code

### DIFF
--- a/backend/functions/src/matching/functions.ts
+++ b/backend/functions/src/matching/functions.ts
@@ -3,7 +3,7 @@ import admin from 'firebase-admin'
 import { FirestoreGroup, Group } from '../types'
 const courseRef = db.collection('courses')
 const studentRef = db.collection('students')
-
+let groupId = 0
 // =====================  HELPER FUNCTIONS ======================
 /** Avoid using this assertion check, it's usually not necessary.
  *  Just get the course doc and throw an error if the data is undefined. */
@@ -90,8 +90,10 @@ async function makeMatches(courseId: string) {
   const newGroups: Group[] = []
   for (let i = 0; i < groupTriples.length; i += 3) {
     groupCounter += 1
+    groupId += 1
     const newGroup = groupTriples.slice(i, i + 3)
     newGroups.push({
+      groupId: groupId,
       groupNumber: groupCounter + lastGroupNumber,
       members: newGroup,
       createTime: nowDate,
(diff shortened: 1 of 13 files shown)


JustKong13:
The changes you've made look really nice! Very clean and helpful change so that we can have easily writable and memorable, unique IDs. The only thing is that you seem to have conflicts with the GroupCard file. It doesn't look like much but if you could get that done it'll be great!

backend/functions/src/matching/functions.ts
```diff
@@ -91,7 +90,9 @@ async function makeMatches(courseId: string) {
   for (let i = 0; i < groupTriples.length; i += 3) {
     groupCounter += 1
     const newGroup = groupTriples.slice(i, i + 3)
+
     newGroups.push({
+      groupId: courseId + groupCounter + lastGroupNumber,
```
mluo24, line 95: Why do you add the `lastGroupNumber` here? The `courseId` and `groupCounter` should be enough to guarantee uniqueness.

mluo24 (approved):
Thank you for working on this Jessica! The solution of generating a unique idea is pretty neat :) I do believe you need to include this in emails as well, which can probably be appended to the end of the template if it's a group, and we'll see what LSC says (rather than editing the templates themselves). You can choose to do this in a new PR or this one, but just remember that it's still a task.

Approving for now.

backend/functions/src/matching/functions.ts
Coreball: Should `groupCounter` even be used? That's a local variable, wouldn't it be reset every time they make new matches, with the `courseId` staying the same?

backend/functions/src/matching/functions.ts
```diff
@@ -104,6 +105,7 @@ async function makeMatches(courseId: string) {
     groupCounter += 1
     const newGroup = groupDoubles.slice(i, i + 2)
     newGroups.push({
+      groupId: courseId + groupCounter + lastGroupNumber,
```
mluo24, line 108: same here

backend/functions/src/matching/functions.ts
mluo24: oh I didn't realize that was the local variable. there should be some what of indicating which group it is lol

Coreball: Actually I see how `groupCounter` and `lastGroupNumber` work together to represent this group's current number, since `lastGroupNumber` isn't incremented when new groups are added. Could be clearer if these were just combined into one thing

PR #160, opened 2023-03-23
Flag a course

b3njamint:
### Current Status 

Can flag and unflag course at free will. Changing pages and jumping around in history will not cause problems.

### Summary 

This pull request allows LSC to flag a course when there is a "bad situation," or when there are potential matches for a course but they are not ideal (i.e. there is only one grad student and one freshmen waiting to be matched. Such a flag would remind LSC to not click the match button.

- [x] implemented X
- [ ] fixed Y

### Test Plan 

The dashboard. You can see the little bookmark icons (could be changed later).
Screen Shot 2023-04-16 at 9 15 59 PM
The emulator. As you can see, there is no flag field right now. Note that this is important, because even if there is no flag field, there are no null problems.
Screen Shot 2023-04-16 at 9 16 32 PM
Oh no, a bad situation!
Screen Shot 2023-04-16 at 9 17 05 PM
Let's flag it.
Screen Shot 2023-04-16 at 9 17 14 PM
A new field is created in the Firebase, and set accordingly. We can refresh, move back and forth in history and the bookmark will still be in its correct state.
Screen Shot 2023-04-16 at 9 20 18 PM

### Notes 

Next steps (look out for a future PR):
- [ ] remove flag when new student joins
- [ ] maybe add some filter option to filter for flags?

### Breaking Changes 

- Changed the CourseContext (added a field to Firebase). I don't see any problems through testing but please let me know.

### DIFF
--- a/backend/functions/src/course/routes.ts
+++ b/backend/functions/src/course/routes.ts
@@ -52,6 +52,9 @@ router.post('/flagged', (req, res) => {
     .catch((err) => {
       logger.error(`Unexpected error updating flagged status: ${err.message}`)
       res.status(500).send({ message: err.message })
+    })
+})
+
 router.get('/semester/current', (_, res) => {
   getCurrentSemester()
     .then((data) => res.status(200).send(data))
--- a/backend/functions/src/course/routes.ts
+++ b/backend/functions/src/course/routes.ts
@@ -1,7 +1,12 @@
 import { Router } from 'express'
 import { config } from 'dotenv'
 import { logger } from 'firebase-functions'
-import { getCourseInfo, getAllCourses, getStudentsForCourse } from './functions'
+import {
+  getCourseInfo,
+  getAllCourses,
+  getStudentsForCourse,
+  setFlagged,
+} from './functions'
 
 const router = Router()
 config()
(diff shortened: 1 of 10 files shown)


rgu0114 (commented):
Nice work Benji! I like how you separated out all your functions to flag a course and used defensive programming to set the flagged field of CourseCard. Everything looks good for the most part, but it would be great if you could describe your testing plan a bit more.

b3njamint:
> Nice work Benji! I like how you separated out all your functions to flag a course and used defensive programming to set the flagged field of CourseCard. Everything looks good for the most part, but it would be great if you could describe your testing plan a bit more.

Thanks for the feedback. I haven't updated the PR description yet, but will do that right now.

mluo24 (approved):
Good job Benji! I like your integration with the CourseContext and the elegant solution to this! I played around with it a bit and it works well.

I'm looking forward to your second PR :)

backend/functions/src/course/routes.ts
```diff
@@ -42,6 +43,15 @@ router.get('/students/:courseId', (req, res) => {
...
 
+// Updates the flagged status in the database
+router.post('/flagged', (req, res) => {
+  const courseId = req.body.courseId
+  const flag = req.body.flagged
+  setFlagged(courseId, flag)
+    .then((data) => res.status(200).send({ flagged: flag }))
+    .catch((err) => {
+      logger.error(`Unexpected error updating flagged status: ${err.message}`)
+      res.status(500).send({ message: err.message })
```
mluo24, line 54: You're missing braces and parentheses at the end.

mluo24 (changes requested):
wait I forgot to say that something went wrong when I ran it because it was a syntax error. I pointed it out in the code.

mluo24 (approved):
works now! good job!

PR #161, opened 2023-03-27
Style settings

JustKong13:
### Summary 

The parts from the old iteration of the settings page have been refactored to follow the new designs, from https://github.com/cornell-dti/zing-lsc/pull/148. In addition to refactoring the old settings page, there is also a section for all the allowed users who have admin permissions. 

- [x] styled page to the designs
- [x] added administrators section 
- [x] ability to delete administrators
- [ ] ability to add administrators
- [ ] ability to edit administrators   

### Test Plan 
Here is what the current settings page looks like: 
image

Let's say we decide to delete user bt283, it turns orange, which serves as a confirmation.
image

Clicking it again will delete user bt283
image

### Notes 

This PR builds off of the branch `update-allowed-users` (https://github.com/cornell-dti/zing-lsc/compare/main...update-allowed-users) which a PR hasn't been made for yet, so this PR is still subject to change in the near future.


rgu0114 (commented):
Nice work Juzzin. I really like the way you separated out the logic for your functions on the Settings page, and the dropdown looks rly good (was very helpful to look at for my task). However, I think it would be best if there was a default value for the dropdown so that it's not just blank when no semester has been selected. Otherwise, LGTM!

jewang25 (commented):
Nice job Justin! It looks good and I like how you integrated the backend and frontend together pretty smoothly. I agree with Richard that it might be nice to add a default value like the current semester or something like that for the dropdown select. This page will be nice to use when we have to add small features that change small things.

frontend/src/modules/Settings/Components/Settings.tsx
```diff
@@ -1,24 +1,87 @@
...
     setCurrRoster(event.target.value)
   }
+
+  const [semesters, setSemesters] = useState<string[]>([])
+  function getAllSemesters() {
+    axios
+      .get(`${API_ROOT}${COURSE_API}/semester/all`)
+      .then((res) => setSemesters(res.data))
+  }
+  getAllSemesters()
```
wizhaaa: I believe this should be wrapped in a use Effect hook. Otherwise, it will continuously call this function anytime this component is mounted/rendered.

frontend/src/modules/Settings/Components/Settings.tsx
```diff
@@ -1,24 +1,87 @@
...
+      })
+      .catch((err) => console.log(err))
+  }
+
+  // TODO: allow to edit admin information
+  const editAdmin = () => {}
+
+  // grabs the state of the backend once when you refresh code
+  const [start, setStart] = useState<boolean>(true)
+  if (start) {
```
wizhaaa: Instead of doing this, it would be better to also put this in an useEffect hook.

frontend/src/modules/Settings/Components/Settings.tsx
```diff
@@ -62,20 +125,82 @@ export const Settings = () => {
...
+            <DropdownSelect
+              value={currRoster}
+              onChange={changeCurrRoster}
+              sx={{
+                alignContent: 'right',
+                right: '1px',
+              }}
+            >
+              {semesters.map((sem) => (
+                <MenuItem value={sem}>{sem}</MenuItem>
```
wizhaaa: MenuItem should have a key prop. Easiest way to fix this is to do: 

```
semesters.map( (sem, i) => { <MenuItem key={i}> ...

```

wizhaaa (changes requested):
Here are some preliminary issues I see with the code - I still have to fully test the functionality of how the survey closes. But there are some issues with controlling the semester-related and survey initial states.

rgu0114:
very good!

JustKong13:
There are still tasks that need to be done, but will and I decided to put that into a separate PR

wizhaaa (approved):

PR #168, opened 2023-04-14
Update Allowed Users

CollinWoo:
### Summary 

This pull request implements backend endpoints for getting and setting the list of allowed users in the Firebase database

- [x] implemented endpoints for adding, removing, and getting the list of allowed users

### Test Plan 

Tests were performed by hitting the backend endpoints in Insomnia and verifying that the responses and updates to the Firebase Database are correct.

### DIFF
--- a/backend/functions/src/index.ts
+++ b/backend/functions/src/index.ts
@@ -9,6 +9,8 @@ import {
   logReqBody,
 } from './middleware/auth-middleware'
 import { unless } from './middleware/unless'
+import { db } from './config'
+import { logger } from 'firebase-functions'
 
 // import routers
 import studentsRouter from './student/routes'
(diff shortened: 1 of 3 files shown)


JustKong13 (commented):
Hi colin! I was working off your branch prior to this PR being made and as of right now I find no issues, every endpoint works exactly as indented. The code is also written extremely efficiently and easy to understand and it was super easy to build off of.

jewang25 (commented):
Nice job Colin! The new functions are really readable and concisely reflect the needed functions of adding, removing, and deleting functions. I like how you added meaningful error messages to each one. It will be useful for the settings page to add, remove, and get users.

wizhaaa (commented):
I think we may want to refactor this so that these functions are in a utils folder rather than directly in the `index.tx` folder => and then import into this file.

mluo24:
> I think we may want to refactor this so that these functions are in a utils folder rather than directly in the `index.tx` folder => and then import into this file.

btw Richard is doing this in his new PR

PR #170, opened 2023-04-17
Flag a course II

b3njamint:
### Summary 

This pull request builds upon #160, where admin can flag courses that they think there is a "bad situation" in. It implements the features mentioned in the previous PR, as well as changing the bookmark icon to a flag.

- [x] automatically remove flag from course when a new student joins
- [x] add filter option to filter for unflagged courses

### Test Plan 

Here, we have two courses with bad situations, so they are flagged.
Screen Shot 2023-04-28 at 12 44 58 PM

I tried adding new students in different flagged/unflagged course combinations.
- just one flagged course
- just one unflagged course (false flagged field)
- just one unflagged course (null flagged field)
- one flagged and one unflagged
- two flagged
- multiple students joining the same flagged course
 
All of these worked, unflagging courses only as necessary.

I also tried the unflagged filter, and it worked too, even when I changed pages and such.

### Notes 

- If you add a student with a different name but the same email, the course doesn't unflag because its probably not another "unique" student?

### Breaking Changes 

- Should be none.


jewang25 (commented):
Nice job Benji! I like how you addressed this task by simply adding another field to the courses in the firebase. I'm not sure if the console.log in the survey file is needed, but if it is then this looks good. I tested it and it seems to work on my end as well.

CollinWoo:
Great work Benji! I tested out the same test cases you did and the code ran perfectly on my end. I especially like how readable and concise your code is. However, I agree with Jessica that the console.log statements in the survey component should be removed.

mluo24 (approved):
Good work Benji! Thanks for taking care of this task!

PR #171, opened 2023-04-24
Settings Patch v0.1 

wizhaaa:
# Summary 

This PR addresses some bug fixes for the settings page and styling updates. 

- [x] Creates a 'global' router that does not require authorization so that student pages can access backend endpoints necessary for functionality. 

> 
> ### Example: 
> the ability to hit the get survey status endpoint (there is no logged-in user, hence no auth) 
> ### Note: 
> I named it global here for the sake of getting something pushed out, I was considering something like `public` (can be confused with the standard public folder for frontend assets) or `utils` (already used). 
> Could change this: because also Richard's PR is also refactoring the backend folders and functions 
> 

- [x] Adds Settings to the account menu dropdown 
- [x] Initializes states correctly now 
- [x] updates the backend state as well; not just local 
- [x] Clarify semester viewing vs. settings changing global 
- [x] Styling enhancement for the survey closed 
- [x] styling for the dashboard a bit so no overflow 

# Notes
I kind of don't like the toggling of survey status - makes it too easy to change something that is kind of important/shouldn't be easily changed. I feel like it should have a security mechanism - some confirmation module or a drop-down or something else - could change this in a future PR. 

We also would have to train/inform/document to LSC on how the different semester switching things work (global vs local views).

Things that may need a change in the future: 
- global router be refactored 
- settings page styling 
- how we implement survey status toggling 
- clarify switching semester view vs. switching semester globally 

# Testing 
https://user-images.githubusercontent.com/46132945/234050809-8f078a07-60ec-46d2-9069-3e5abe268849.mov

### New Settings Page Preview: 
image

### DIFF
--- a/backend/functions/src/course/functions.ts
+++ b/backend/functions/src/course/functions.ts
@@ -16,7 +16,7 @@ export const setCurrentSemester = async (sem: String) => {
 
 export const getAllSemesters = async () => {
   const semData = (await semesterRef.get()).data() as Semester
-  return [semData.currentSemester, ...semData.allSemesters]
+  return semData.allSemesters
 }
 
 export const getSurveyStatus = async () => {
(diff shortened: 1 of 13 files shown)


mluo24:
Will please do code review

mluo24 (approved):
good work team

now we need the devs to fix all of our changes

PR #164, opened 2023-03-30
No match session closed 2

b3njamint:
### Summary 

This pull request is the revamped version of No match, session closed :( #142.
It adds an email template for students who do not have a match after about a 6-week period, which the LSC users would keep track of manually. 

- [ ] implemented X
- [x] fixed Y

### Test Plan 

I tested in the staging app: https://zing-lsc.web.app/

The template shows up!
Screen Shot 2023-03-30 at 4 44 03 PM

The contents are correctly filled in as well.
Screen Shot 2023-03-30 at 4 44 09 PM

### Notes 

Will added the template to prod, so that part is taken care of.

### Breaking Changes 

- Added template, so Firebase is modified.

PR #173, opened 2023-04-24
No match session closed 3

b3njamint:
### Summary 

This pull request is the exact same as No match, session closed 2 #164, but there were some snowballing merge conflicts in the previous PR.
It adds an email template for students who do not have a match after about a 6-week period, which the LSC users would keep track of manually. 

- [ ] implemented X
- [x] fixed Y

### Test Plan 

I tested in the staging app: https://zing-lsc.web.app/

The template shows up!
Screen Shot 2023-03-30 at 4 44 03 PM

The contents are correctly filled in as well.
Screen Shot 2023-03-30 at 4 44 09 PM

### Notes 

Will added the template to prod, so that part is taken care of.

### Breaking Changes 

- Added template, so Firebase is modified.


mluo24 (approved):
Nice work Benji! I'm glad we can finally merge this in.

PR #174, opened 2023-04-24
Added groupId to Emails

jewang25:
### Summary 

Added groupId as the last line of an email by LSC so that LSC knows which group they are emailing. 

### Test Plan 
Tested that the right groupId showed up for each group and that they were different. Tested that the groupId does not show up for the no match templates and the individual emails.

Screenshot 2023-04-23 at 1 51 37 PM
Screenshot 2023-04-23 at 1 54 40 PM


mluo24 (approved):
Thank you for this simple fix Jessica! It works for when you send group emails automatically :)

PR #176, opened 2023-04-25
Add admin

JustKong13:
### Summary 

This PR is the first step in making the full functionality to add admins to the allowed users list

- [x] implemented ability to add admins (name and admins)
- [x] shows the new admin on the newly updated administrator table

### Test Plan 

The simple act of being able to add an admin is there 

In this example, I was originally deleted
image

And then I filled myself out as a new administrator
image

After adding myself, I appear back on the administrators' table
image

### Notes 

Edit function still does not work but that can be overrided by removing an admin and then adding the same admin with updated information


b3njamint (commented):
Very nice fix! I tried playing around with it and didn't find anything bad. I really like how you had the undo button in the "confirmation state," but it's really big and stretches the table row (idk if that is intentional).

Is the edit feature going to be done in a future PR?

mluo24 (approved):
Nice work Justin! I would just remove the edit icons and you should be good to go. I like how you accounted for duplicate emails.

mluo24:
fixed it up and merged main.

PR #178, opened 2023-04-26
Fix switch semester endpoint

CollinWoo:
### Summary 

This pull request adds requested fixes to the switch semester endpoint

- [x] switching the current semester adds the previous current semester to the list of all semesters
- [x] switching the current semester removes the new current semester from the list of all semesters

### Test Plan 

Tests were done in Insomnia to make sure the functionality works as intended.


JustKong13 (commented):
This endpoint looks good and resolves the issue back then with the settings where there was no way that when you added a semester it would also add to the list of semesters. However, I think this already exists in prod so it's not really that necessary at this point in time.

rgu0114:
Great work on this task Colin, and thank you for working on it. The endpoint looks good and I like how you were able to update the current semester properly with minimal changes/very cleanly.

mluo24:
Functionality is already in main. Thanks Colin!

PR #180, opened 2023-05-08
Added edge case for no match session closed

jewang25:
### Summary 

- added edge case for no match session closed filter
- checked if the template timestamp key no-match-session-closed was undefined or not and added it as a condition to the studentHasUnsentNoMatch function

### Test Plan 

- tested sending both no-match-session-closed and no-match emails
- tested filtering on no no match emails and other filters


mluo24 (approved):
Thanks Jessica for a simple and quick solution! Looks good to me :)

PR #182, opened 2023-07-29
Survey uses current semester setting instead of a hardcoded default

mluo24:
### Summary 

LSC reported that despite changing the semester, SP23 was still showing. This was because in the backend, the roster defaulted to a hardcoded string rather than the dynamic setting we now have. This PR fixes that.

### Test Plan 

Local run to change semester, then submit survey to see if issue is still there.

image


wizhaaa (approved):
Thanks for the quick fix!

PR #185, opened 2024-01-03
Bug fixes for wi24

JustKong13:
### Summary 

This pull request fixes a few bugs and prepares LSC for the winter semester, as well as 
- [x] Dashboard no longer preloads spring 23
- [x] Fixed Email Regex to have to end with `cornell.edu`
- [x] Added winter '24 semester to metrics 

### Test Plan 

Metrics page also has winter '24 as a valid option
Screenshot 2024-01-03 at 5 16 22 PM

Now correctly identifies invalid emails
Screenshot 2024-01-03 at 5 15 37 PM

### Notes 

In the future we will need to make the metrics semesters update automatically. 

### Breaking Changes 

`yarn.lock` was updated when I did `yarn install`, it seems to me that the changes primarily consists of a new syntax update which just removed quotation marks. 
`package-lock.json` was also updated when I did `npm install`, due to me working off of a new machine. There were no deletions or additions, just standard upgrades.

### DIFF
--- a/backend/functions/src/student/functions.ts
+++ b/backend/functions/src/student/functions.ts
@@ -112,7 +112,7 @@ export const addStudentSurveyResponse = async (
   }
 
   // 0. Check if email is valid cornell.edu email.
-  const emailRegex = /^\w+@cornell.edu$/
+  const emailRegex = /^\w+@cornell\.edu$/
   // if true => valid email. if false => invalid email.
   const validEmail = emailRegex.test(email)
   if (!validEmail) {
--- a/backend/functions/package-lock.json
+++ b/backend/functions/package-lock.json
@@ -36,9 +36,9 @@
       }
     },
     "node_modules/@babel/parser": {
-      "version": "7.21.4",
-      "resolved": "https://registry.npmjs.org/@babel/parser/-/parser-7.21.4.tgz",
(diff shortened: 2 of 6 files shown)


backend/functions/src/student/functions.ts
```diff
@@ -112,7 +112,7 @@ export const addStudentSurveyResponse = async (
   }
 
   // 0. Check if email is valid cornell.edu email.
-  const emailRegex = /^\w+@cornell.edu$/
+  const emailRegex = /^\w+@cornell\.edu$/
```
rgu0114, line 115: idr how regex works oop, does this like, make sure emails end with explicitly .edu and not like .education?

rgu0114 (commented):
Maybe mention what dependency changes u made in package.lock/yarn.lock

backend/functions/src/student/functions.ts
Coreball: Probably has to do with the dot matching any character other than line terminators lol

rgu0114: ah ok. btw, hi Changyuan!

JustKong13:
> Maybe mention what dependency changes u made in package.lock/yarn.lock

JustKong13:
> Maybe mention what dependency changes u made in package.lock/yarn.lock

Just added that to the pull request! Basically I think for frontend there was an update which removed quotation marks around certain parts of the file (with no meaningful dependency changes), and for the backend our packages had version upgrades. There was no addition or deletion of any packages

rgu0114 (approved):
looks good 👍

PR #186, opened 2024-02-13
Metrics now automatically update with new semesters

JustKong13:
### Summary 

Previously when we moved to a new semester, someone needed to manually change the codebase in order to make that semester available to be viewed on the metrics page. 

- [x] Newly added semesters automatically also get added to metrics
- [x] Metrics page automatically loads with the current semester set by in the settings  

### Test Plan 

Originally on the metrics page there isn't an option for Spring 25. 
Screenshot 2024-02-15 at 12 26 11 AM

I can now add the Spring 25 semester via the settings. 
Screenshot 2024-02-15 at 12 27 03 AM

And now it is available as an option. 
Screenshot 2024-02-15 at 12 29 06 AM

### Notes 

The dropdown menu of the metrics page has the abbreviation semester "SP", rather than the the fully spelled out word "spring". 

### Breaking Changes

### DIFF
--- a/frontend/src/modules/Metrics/Components/Metrics.tsx
+++ b/frontend/src/modules/Metrics/Components/Metrics.tsx
@@ -12,9 +12,11 @@ import { DASHBOARD_PATH } from '@core/Constants'
 import { Link } from 'react-router-dom'
 import { GroupMembership } from '@core'
 import survey from '@core/Questions/Questions.json'
+import { useSettingsValue } from '@context/SettingsContext'
 
 export const Metrics = () => {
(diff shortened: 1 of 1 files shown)


rgu0114 (approved):
Great work on this PR Justin, thanks for continuing to make the Zing codebase extensible and more self-sufficient for the future! The functionality looks good – the only thing I would note is that admin users are able to enter negative numbers for the year, but that's okay for now/can be addressed in a future PR. Overall though, this looks great and I liked the simplicity in your solution.

PR #187, opened 2024-03-05
fixed check-in filter on frontend

JustKong13:
### Summary 

This pull request fixes the bug that the "no check-in" filter was showing classes in which all groups were already sent a check-in email. 

This bug stemmed from the fact that in order to display the order of email templates accurately in the send email modal, the backend fields were updated with numbers to reflect that order of email templates alphanumerically, rather than fixing the order on the frontend. Although this field name was changed in the database, it was not reflected in the filter logic. In this PR, filter logic was then updated to reflect the updated field. 

Screenshot 2024-03-05 at 1 50 03 PM

### Test Plan 

In this example, ECON 1110 and PHYS 2213 students were matched, and updated the `templateTimestamps` field in the emulator database through the backend for ECON 1100. This was done through calling the `updateGroupTimestamp` in `backend/functions/src/emailing/functions.ts`.

Since the Microsoft Graph PR (https://github.com/cornell-dti/zing-lsc/pull/165) was not merged, I was not able to use the emailing feature via the frontend, which is why I called the backend function instead so that the database can update appropriately. 

Screenshot 2024-03-05 at 1 29 14 PM

After applying the filter, ECON 1100 no longer shows up when applying the "no checkin email" filter, as expected.

Screenshot 2024-03-05 at 1 38 35 PM

### Notes 

This change has no affect on the ability of the system to get email templates. Although the id for the check-in email is `check-in`, it is still able to retrieve the correct email template since the filter does not communicate with the email templates.

### Breaking Changes

### DIFF
--- a/frontend/src/modules/Dashboard/Components/Dashboard.tsx
+++ b/frontend/src/modules/Dashboard/Components/Dashboard.tsx
@@ -70,7 +70,7 @@ export const Dashboard = () => {
   )
   //Helper function to check if a given course has any groups without check-in emails
   function hasUnsentCheckIns(c: Course) {
-    return c.groups.some((group) => !group.templateTimestamps['check-in'])
+    return c.groups.some((group) => !group.templateTimestamps['1check-in'])
   }
 
   //Helper function that returns true if the student doesn't have a no match email


rgu0114 (approved):
Looks good Justin, and thanks for the detailed explanation of your changes! It was definitely helpful in understanding the reasoning behind the change.

